PR #9, opened 2020-06-26
async memfs and async fuse

pwang7:
I tested this async memfs on Linux with ARM64.
The basic filesystem functionalities work, but not quiet stable yet.
I also made some fixes to async fuse.

### DIFF
--- a/async_fuse/src/fs/mod.rs
+++ b/async_fuse/src/fs/mod.rs
@@ -21,7 +21,7 @@ mod util;
 use dir::*;
 use node::*;
 
-const MY_TTL_SEC: u64 = 1; // TODO: should be a long value, say 1 hour
+const MY_TTL_SEC: u64 = 3600; // TODO: should be a long value, say 1 hour
 const MY_GENERATION: u64 = 1; // TODO: find a proper way to set generation
 
 #[derive(Debug)]
@@ -59,7 +59,7 @@ impl FileSystem {
                 occupied.ino(),
             );
             reply.error(EEXIST).await?;
-            return Err(anyhow::anyhow!("cannot create duplicated node"));
+            return Ok(());
         }
         // all checks are passed, ready to create new node
         let mflags = util::parse_mode(mode);
@@ -214,9 +214,7 @@ impl FileSystem {
                         node_name, parent,
                     );
                     reply.error(ENOENT).await?;
-                    return Err(anyhow::anyhow!(
-                        "cannot find the node to remove under parent directory"
-                    ));
+                    return Ok(());
                 }
                 Some(child_entry) => {
                     node_ino = child_entry.ino();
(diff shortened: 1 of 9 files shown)


async_fuse/src/fs/dir.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    #[test]
+    fn test_dir() -> nix::Result<()> {
+        smol::run(async {
+            let oflags = OFlag::O_RDONLY | OFlag::O_DIRECTORY;
+            let fd = blocking!(fcntl::open(".", oflags, Mode::empty()))?;
+            let dir = Dir::from_fd(fd)?;
+            let mut dir = smol::iter(dir);
+
+            while let Some(entry) = dir.next().await {
```
rogercloud, line 137: next is a regular function, should we await it?

async_fuse/src/fs/mod.rs
```diff
@@ -0,0 +1,1420 @@
...
+        cache.insert(FUSE_ROOT_ID, root_inode);
+        let trash = BTreeSet::new(); // for deferred deletion
+
+        Ok(FileSystem { cache, trash })
+    }
+
+    /// Initialize filesystem.
+    /// Called before any other filesystem method.
+    pub fn init(&mut self, _req: &Request<'_>) -> anyhow::Result<()> {
+        Ok(())
```
rogercloud, line 283: Is this function empty on purpose?

async_fuse/src/fs/mod.rs
```diff
@@ -0,0 +1,1420 @@
...
+    }
+
+    /// Clean up filesystem.
+    /// Called on filesystem exit.
+    pub fn destroy(&mut self, _req: &Request<'_>) {}
```
rogercloud, line 288: the same comment as above.

async_fuse/src/fs/util.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub rdev: u32,
+    /// Flags (macOS only, see chflags(2))
+    pub flags: u32,
+}
+
+pub fn parse_oflag(flags: u32) -> OFlag {
+    debug_assert!(
+        flags < std::i32::MAX as u32,
+        format!(
+            "helper_parse_oflag() found flags={} overflow, larger than u16::MAX",
```
rogercloud, line 49: the log should be "larger than i32::MAX"

async_fuse/src/fs/dir.rs
```diff
@@ -0,0 +1,150 @@
...
+#[cfg(not(target_os = "linux"))]
+use libc::{dirent, readdir_r};
+
+#[derive(Debug)]
+pub(crate) struct Dir(ptr::NonNull<libc::DIR>);
+
+impl Dir {
+    /// Converts from a file descriptor, closing it on success or failure.
+    pub fn from_fd(fd: RawFd) -> nix::Result<Self> {
+        let d = unsafe { libc::fdopendir(fd) };
```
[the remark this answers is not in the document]
pwang7: I did use nix::Dir before, but nix::Dir::iter() uses Dir as borrowed reference, which is not friendly for async.
That's why I implement Dir by myself.

async_fuse/src/fs/mod.rs
pwang7: This function can be removed.

Issue #49, opened 2020-07-09
Clippy Fix 3


rogercloud:
Fix following clippy warning
```
warning: method is never used: `statfs`
   --> async_fuse/src/fuse_reply.rs:399:5
    |
399 | /     pub async fn statfs(
400 | |         self,
401 | |         blocks: u64,
402 | |         bfree: u64,
...   |
425 | |             .await
426 | |     }
    | |_____^

warning: method is never used: `created`
   --> async_fuse/src/fuse_reply.rs:446:5
    |
446 | /     pub async fn created(
447 | |         self,
448 | |         ttl: &Duration,
449 | |         attr: FuseAttr,
...   |
471 | |             .await
472 | |     }
    | |_____^

warning: method is never used: `locked`
   --> async_fuse/src/fuse_reply.rs:492:5
    |
492 |     pub async fn locked(self, start: u64, end: u64, typ: u32, pid: u32) -> anyhow::Result<()> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: this function has too many arguments (15/7)
   --> fuse_ll/src/fuse/mod.rs:129:5
    |
129 | /     fn setattr(
130 | |         &mut self,
131 | |         _req: &Request<'_>,
132 | |         _ino: u64,
...   |
144 | |         reply: ReplyAttr,
145 | |     ) {
    | |______^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments

warning: method is never used: `bmap`
   --> async_fuse/src/fuse_reply.rs:523:5
    |
523 |     pub async fn bmap(self, block: u64) -> anyhow::Result<()> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: method is never used: `error`
   --> async_fuse/src/fuse_reply.rs:582:5
    |
582 |     pub async fn error(self, err: c_int) -> anyhow::Result<()> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: method is never used: `size`
   --> async_fuse/src/fuse_reply.rs:599:5
    |
[...]

PR #62, opened 2020-07-12
#49 Clippy Fix 3

pzheng1025:
Ignore unused codes.
Skip to fix too many arguments.


async_fuse/src/fuse_request.rs
```diff
@@ -826,4 +826,10 @@ mod tests {
             _ => panic!("unexpected request operation"),
         }
     }
+
+    #[test]
```
rogercloud, line 830: What's this test for?

pzheng1025: > What's this test for?
To fix this, it is easy to add a test case, so I don't use #[allow(dead_code)]
warning: method is never used: `len`
  --> async_fuse/src/fuse_request.rs:17:5
   |
17 |     pub fn len(&self) -> usize {
   |     ^^^^^^^^^^^^^^^

pwang7: Will test remove clippy unused warning?

pzheng1025: Just tested. It won't. Will submit a new one.

pwang7:
Please update branch before merge

PR #70, opened 2020-07-14
add buffer pool to async fuse

pwang7:
use crossbeam bounded channel as lock-free buffer pool.


async_fuse/src/session.rs
```diff
@@ -190,9 +202,13 @@ impl Session {
                             // TODO: this panic is for fast fail, can be removed when stable
                             panic!("failed to process request, the error is: {}", e);
                         }
-                        byte_vec
+                        let res = sender.send((idx, byte_vec));
```
rogercloud, line 205: should we reset the memory to 0s before send it back?

pwang7: I think it should be fine without memset the buffer to zero.

I fixed some bugs, now the test should pass.

rogercloud: byte_vec length should be reset to its max value.

Issue #51, opened 2020-07-09
Clippy Fix 5


rogercloud:
Fix following clippy warning
```
warning: constant item is never used: `FUSE_MIN_READ_BUFFER`
   --> async_fuse/src/protocal.rs:426:1
    |
426 | pub const FUSE_MIN_READ_BUFFER: usize = 8192;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: this function has too many arguments (10/7)
   --> fuse_ll/src/fuse/mod.rs:468:5
    |
468 | /     fn getlk(
469 | |         &mut self,
470 | |         _req: &Request<'_>,
471 | |         _ino: u64,
...   |
478 | |         reply: ReplyLock,
479 | |     ) {
    | |______^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments

warning: constant item is never used: `FUSE_COMPAT_STATFS_SIZE`
   --> async_fuse/src/protocal.rs:705:1
    |
705 | pub const FUSE_COMPAT_STATFS_SIZE: usize = 48;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: method is never used: `fd`
  --> async_fuse/src/session.rs:82:5
   |
82 |     pub fn fd(&self) -> RawFd {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^

warning: this function has too many arguments (11/7)
   --> fuse_ll/src/fuse/mod.rs:490:5
    |
490 | /     fn setlk(
491 | |         &mut self,
492 | |         _req: &Request<'_>,
493 | |         _ino: u64,
...   |
501 | |         reply: ReplyEmpty,
502 | |     ) {
    | |______^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments

warning: this function has too many arguments (15/7)
   --> async_fuse/src/fs/mod.rs:529:5
    |
529 | /     pub async fn setattr(
530 | |         &mut self,
531 | |         req: &Request<'_>,
532 | |         ino: u64,
...   |
544 | |         reply: ReplyAttr,
545 | |     ) -> anyhow::Result<()> {
    | |___________________________^
    |
    = note: `#[warn(clippy::too_many_arguments)]` on by default
[...]

PR #72, opened 2020-07-14
#51 Clippy Fix 5

MrWWheat:
Ignore unused codes.
Split map and count.
Skip to fix too many arguments.

### DIFF
--- a/async_fuse/src/protocol.rs
+++ b/async_fuse/src/protocol.rs
@@ -141,120 +141,175 @@ pub const FATTR_BKUPTIME: u32 = 1 << 30;
 #[cfg(target_os = "macos")]
 pub const FATTR_FLAGS: u32 = 1 << 31;
 
-// Flags returned by the OPEN request
-//
-// FOPEN_DIRECT_IO: bypass page cache for this open file
-// FOPEN_KEEP_CACHE: don't invalidate the data cache on open
-// FOPEN_NONSEEKABLE: the file is not seekable
-// FOPEN_CACHE_DIR: allow caching this directory
-// FOPEN_STREAM: the file is stream-like (no file position at all)
-//
-pub const FOPEN_DIRECT_IO: u32 = 1 << 0;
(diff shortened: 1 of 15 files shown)


async_fuse/src/protocol.rs
```diff
@@ -424,6 +424,7 @@ pub enum FuseNotifyCode {
 }
 
 // The read buffer is required to be at least 8k, but may be much larger
+#[allow(dead_code)]
```
pwang7, line 427: for unused const variables, add a mod to group them, and allow dead_code for the mod.

fuse_ll/src/memfs.rs
```diff
@@ -4,6 +4,7 @@ use crate::fuse::{
 };
 use libc::{EEXIST, EINVAL, ENODATA, ENOENT, ENOTEMPTY};
 use log::{debug, error}; // info, warn
+use nix::dir::Entry;
 use nix::dir::{Dir, Type};
```
pwang7, line 8: use nix::dir::{Dir, Entry, Type};

MrWWheat: done

async_fuse/src/protocol.rs
MrWWheat: done

async_fuse/src/protocol.rs
```diff
@@ -478,8 +478,15 @@ pub enum FuseNotifyCode {
     FUSE_NOTIFY_CODE_MAX,
 }
 
-// The read buffer is required to be at least 8k, but may be much larger
-pub const FUSE_MIN_READ_BUFFER: usize = 8192;
+#[allow(dead_code)]
+pub mod fuse_configs {
```
pwang7, line 482: how about group more related const variables and use a more meaningful mod name?

MrWWheat: done.

PR #74, opened 2020-07-15
deny some lint warnings which are allowed by default

pwang7:
there are many lint warnings are allowed by default, so deny these allowed lint warnings.
some lint warnings need to fixed before deny them.
add some more issue to track them.

### DIFF
--- a/async_fuse/src/main.rs
+++ b/async_fuse/src/main.rs
@@ -1,3 +1,29 @@
+#![deny(
+    // The following are allowed by default lints according to
+    // https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html
+    anonymous_parameters,
+    bare_trait_objects,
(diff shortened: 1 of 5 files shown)


async_fuse/src/main.rs
```diff
@@ -1,3 +1,29 @@
+#![deny(
+    // The following are allowed by default lints according to
+    // https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html
+    anonymous_parameters,
+    bare_trait_objects,
+    // box_pointers, TODO: fix box pointers
```
rogercloud, line 6: What's following plan for these TODO?

pwang7: some are easy to fix, such as unreachable_pub, some need more work, such as unused_results, unsafe_code, missing_docs.

I'll open some issues to track them.

PR #77, opened 2020-07-15
minor fix for buffer pool

### LINKED ISSUE #27: FUSE request buffer pool
To avoid data copy when receiving FUSE request, it needs to hold a buffer for each FUSE request.
When handling FUSE request concurrently, it needs to have multiple buffers, each for one FUSE request exclusively.
When a FUSE request is finished processing, its buffer should be return back to the buffer pool to be reused for next FUSE request.
The buffer pool size can be defined as [the number of max pending requests](https://github.com/libfuse/libfuse/blob/master/include/fuse_common.h#L484).

pwang7:
Close #27

### DIFF
--- a/async_fuse/src/session.rs
+++ b/async_fuse/src/session.rs
@@ -111,7 +111,10 @@ impl Session {
         (0..MAX_BACKGROUND).for_each(|i| {
             let res = pool_sender.send((i, vec![0u8; BUFFER_SIZE]));
             if let Err(e) = res {
-                panic!("failed to insert buffer idx={} to buffer pool", i);
+                panic!(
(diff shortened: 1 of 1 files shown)


async_fuse/src/session.rs
```diff
@@ -157,6 +163,9 @@ impl Session {
                     let fs = self.filesystem.clone();
                     let sender = pool_sender.clone();
                     Task::spawn(async move {
+                        // This task will run in parallel, store the buffer index to avoid data race
+                        let buf_idx = idx;
```
rogercloud, line 167: Why is "idx" not moved inside the async block?

pwang7: you are right, idx will be copied as soon as the task is spawned.

PR #90, opened 2020-07-16
enable cargo audit and cargo outdated in CI

pwang7:


.github/workflows/ci.yml
```diff
@@ -1,8 +1,35 @@
-on: pull_request
-
 name: CI
+on:
+
+  push:
```
rogercloud, line 4: Currently it's not allowed to push master.

pwang7: done

Issue #40, opened 2020-07-04
Add fuse CLI parser and mount option parser


pzheng1025:
Refer to https://gist.github.com/c4milo/2007941

PR #92, opened 2020-07-19
#40 Add mount options parser and options validator

pzheng1025:


fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,158 @@ use std::path::Path;
...
+
+    use super::MountOption;
+    use regex::Regex;
+    // Safe to use unwrap here becuase expression is always valid.
+    pub fn get_mount_options() -> Vec<MountOption<'static>> {
+        vec![
+            MountOption {
+                name: "ro",
+                parser: FuseMountArgs::parse_flag,
+                regex: Regex::new("^ro$").unwrap(),
```
pwang7, line 75: do not use unwrap() directly

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,158 @@ use std::path::Path;
...
+            },
+            MountOption {
+                name: "rw",
+                parser: FuseMountArgs::parse_flag,
+                regex: Regex::new("^rw$").unwrap(),
+                flag: None,
+            },
+            MountOption {
+                name: "allow_other",
+                parser: FuseMountArgs::parse_allow_other,
```
pwang7, line 86: define a closure here, instead of a function

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,158 @@ use std::path::Path;
...
+                regex: Regex::new("^allow_other$").unwrap(),
+                flag: None,
+            },
+            MountOption {
+                name: "fsname=<name>",
+                parser: FuseMountArgs::parse_fsname,
```
pwang7, line 92: define a closure here, instead of a function

fuse_ll/src/fuse/mount.rs
pzheng1025: // Safe to use unwrap here becuase expression is always valid.
I have add comment at line 69. Because the regex is literal string and always valid at compile time. 
So I think it is OK to unwrap here. How do you think?

fuse_ll/src/fuse/mount.rs
pzheng1025: Thanks. Will do.

fuse_ll/src/fuse/mount.rs
pzheng1025: Thanks. Will do.

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,158 @@ use std::path::Path;
 
 use param::*;
 
+pub struct MountOption {
+    pub name: String,
+    // validator: fn(&str) -> bool,
```
pwang7, line 16: remove commented code.

pzheng1025: Fixed in commit#4

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,157 @@ use std::path::Path;
...
+    pub flag: Option<u64>,
+    #[cfg(target_os = "macos")]
+    pub flag: Option<i32>,
+    #[cfg(target_os = "macos")]
+    pub fuse_flag: Option<u64>,
+}
+
+pub fn get_all_options() -> String {
+    let options: Vec<_> = get_mount_options().iter().map(|x| x.name.clone()).collect();
+    options.join(",")
```
rogercloud, line 28: merge into one line?

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,157 @@ use std::path::Path;
...
+}
+
+pub fn options_validator(option: String) -> Result<(), String> {
+    let ops: Vec<_> = option.split(',').collect();
+    let ret = ops
+        .iter()
+        .all(|&op| get_mount_options().iter().any(|x| x.regex.is_match(&op)));
```
rogercloud, line 35: ops is used nowhere else. Could we merge them into one statement?

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,157 @@ use std::path::Path;
...
+        fn parse_fsname(args: &mut FuseMountArgs, _mount_option: &MountOption, option: &str) {
+            let name = String::from(option.split('=').last().unwrap()); //Safe to use unwrap here, becuase option is always valid.
+            args.fsname = Some(name);
+            args.fusermount_opts = add_option(&args.fusermount_opts, option);
+        }
+        // Safe to use unwrap here, because expression is literal string and always valid.
+        let ro_regex = Regex::new("^ro$").unwrap();
+        let rw_regex = Regex::new("^rw$").unwrap();
+        let allow_other_regex = Regex::new("^allow_other$").unwrap();
+        let fsname_regex = Regex::new("^fsname=.+$").unwrap();
```
rogercloud, line 86: will white space be included in the regex? ".+" will include all possible characters.

fuse_ll/src/fuse/mount.rs
```diff
@@ -10,21 +11,157 @@ use std::path::Path;
...
+            args.kernel_opts = add_option(&args.kernel_opts, option);
+        }
+
+        fn parse_fsname(args: &mut FuseMountArgs, _mount_option: &MountOption, option: &str) {
+            let name = String::from(option.split('=').last().unwrap()); //Safe to use unwrap here, becuase option is always valid.
+            args.fsname = Some(name);
+            args.fusermount_opts = add_option(&args.fusermount_opts, option);
+        }
+        // Safe to use unwrap here, because expression is literal string and always valid.
+        let ro_regex = Regex::new("^ro$").unwrap();
```
rogercloud, line 83: How to make sure there's no overlap between the regexes? Is there any easier way to match instead of regexes?

fuse_ll/src/fuse/mount.rs
pzheng1025: Rust cannot infer the type of options, so separate to two lines.

fuse_ll/src/fuse/mount.rs
pzheng1025: Same as last one.

fuse_ll/src/fuse/mount.rs
pzheng1025: Will fix.

fuse_ll/src/fuse/mount.rs
pzheng1025: There are two kinds of options. -o aa and -o aa=bb regex is a clean way. Otherwise will use string match.

pzheng1025: Because the mount option only have fixed patterns. -o aa  or -o aa=bb, aa is literal string, so it is impossible to overlap.

fuse_ll/src/fuse/mount.rs
rogercloud: let a = [1, 2, 3];

let doubled = a.iter().map(|x| x * 2).collect::<Vec<i32>>();

fuse_ll/src/fuse/mount.rs
rogercloud: The regex match can't be used in Hashmap, but the string could. The complexity of regex match is N^2, but hashmap version is N, in which N is the total possible options count.

fuse_ll/src/fuse/mount.rs
pzheng1025: Thanks

fuse_ll/src/fuse/mount.rs
pzheng1025: https://github.com/datenlord/datenlord/issues/96
Open an issue for this improvement. Will fix in another PR.

Issue #96, opened 2020-07-23
Improve command line parser performance.


pzheng1025:
**Is your feature request related to a problem? Please describe.**
Current all options are stored in a vector. When options are parsed, the worse case will be O(n^2).

**Describe the solution you'd like**
Use hashmap to store options to reduce to O(n)

PR #98, opened 2020-07-25
#96 Improve command line parser performance.

pzheng1025:
Use hashmap to store mount options to reduce search to O(1)

### DIFF
--- a/fuse_ll/src/fuse/mount.rs
+++ b/fuse_ll/src/fuse/mount.rs
@@ -3,6 +3,7 @@ use nix::errno::{self, Errno};
 use nix::fcntl::{self, OFlag};
 use nix::sys::stat::{self, FileStat, Mode};
 use regex::Regex;
+use std::collections::HashMap;
 use std::ffi::CString;
 use std::fs;
 use std::os::raw::c_void;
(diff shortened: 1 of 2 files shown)


fuse_ll/src/fuse/mount.rs
```diff
@@ -148,13 +157,12 @@ mod param {
...
-                let mount_options = get_mount_options();
-                let option = mount_options
-                    .iter()
-                    .find(|x| x.regex.is_match(&op))
-                    .unwrap();
-                (option.parser)(&mut args, &option, &op)
+            let mount_options_map = super::get_mount_options_map();
+            options.iter().for_each(|op| {
+                let key = op.split('=').collect::<Vec<_>>()[0].to_string();
+                if let Some(option) = mount_options_map.get(&key) {
```
rogercloud, line 163: Will we report error when the option is not defined?

fuse_ll/src/fuse/mount.rs
```diff
@@ -283,13 +289,12 @@ mod param {
...
```
rogercloud, line 295: The same as above

fuse_ll/src/fuse/mount.rs
pzheng1025: All the options that reach here will be valid options.
Option validation will be done by command line parser.

rogercloud: If it's garanteed to be validate, `unwrap` is a clear way I think.

pzheng1025: Modified. Thanks.

PR #100, opened 2020-07-25
Use real mount options for macos.

MrWWheat:
Pass the real mount options to libc::mount.
Initialize the FuseMountArgs when creating the object.


fuse_ll/src/fuse/mount.rs
```diff
@@ -493,7 +514,7 @@ pub fn umount(mount_point: &Path) -> i32 {
 
 #[cfg(any(target_os = "macos"))]
 pub fn mount(mount_point: &Path, options: &[&str]) -> RawFd {
-    let _args = FuseMountArgs::parse(options);
+    let mut _args = FuseMountArgs::parse(options);
```
pzheng1025, line 517: remove '_',  the original '_' is to suppress unused variable error.

pzheng1025: Remove underscore _

MrWWheat: Done.

fuse_ll/src/fuse/mount.rs
```diff
@@ -543,54 +564,25 @@ pub fn mount(mount_point: &Path, options: &[&str]) -> RawFd {
...
-        fsid: 0,
-        fssubtype: FUSE_FSSUBTYPE_UNKNOWN,
-        iosize: FUSE_DEFAULT_IOSIZE,
-        random: drandom,
-        rdev: sb.st_rdev as u32,
-    };
+
+    args.set_mntpath(mntpath_slice);
+    args.set_random(drandom);
+    args.set_rdev(sb.st_rdev as u32);
```
pwang7, line 573: as conversion will trigger clippy lint warning, consider implement From trait for conversion

MrWWheat: I will make another PR to implement the From trait, since this PR is only supposed to use real mount options for macos.

Issue #8, opened 2020-06-23
Lock-free cache manage system


rogercloud:
We manage cache in the memfs, and the cache is supposed to be manipulated by multiple threads simultaneously. The cache manage system is used to handle parallel access. The lock-free data structure is one of the options. The finalized decision is made based on benchmark and profiling.

francis0407:
This is a proposal for an implementation of lockfree hashmap with cuckoo hashing scheme.

## Introduction

Cuckoo hashing is an open addressing solution for hash collisions. The basic idea of cuckoo hashing is to resolve collisions by using two or more hash functions instead of only one. In thisimplementation, we use two hash functions and two arrays (or tables).

The search operation only looks up two slots, i.e. table[0][hash0(key)] and table[1][hash1(key)]. If these two slots do not contain the key, the hash table do not contain the key. So the search operation only takes a constant time in the worst case.

The insert operation must pay the price for the quick search. The insert operation can only put the key into one of the two slots. However, when both slots are already full, it will be necessary to move other keys to their second locations (or back to their first locations) to make room for the new key, which is called a `relocation`. If the moved key can't be relocated because the other slot of it is also occupied, another `relocation` is required. If relocation is a very long chain or meets a infinite loop, the table should be resized or rehashed.

## Schedule

- [x]  Implement the basic `search`, `insert`, `remove` methods.

- [ ]  Implement lock-free `resize`.

- [ ]  Implement other optimizations.

## Reference
* Nguyen, N., & Tsigas, P. (2014). Lock-Free Cuckoo Hashing. 2014 IEEE 34th International Conference on Distributed Computing Systems, 627-636. [pdf](https://github.com/datenlord/datenlord/files/5072865/ICDCS14.pdf)

PR #106, opened 2020-08-05
Add a lockfree hashmap

francis0407:
This PR introduces an implementation of lockfree hashmap with cuckoo hashing, as part of https://github.com/datenlord/datenlord/issues/8.

## Introduction

Cuckoo hashing is an open addressing solution for hash collisions. The basic idea of cuckoo hashing is to resolve collisions by using two or more hash functions instead of only one. In thisimplementation, we use two hash functions and two arrays (or tables).

The search operation only looks up two slots, i.e. table[0][hash0(key)] and table[1][hash1(key)]. If these two slots do not contain the key, the hash table do not contain the key. So the search operation only takes a constant time in the worst case.

The insert operation must pay the price for the quick search. The insert operation can only put the key into one of the two slots. However, when both slots are already full, it will be necessary to move other keys to their second locations (or back to their first locations) to make room for the new key, which is called a `relocation`. If the moved key can't be relocated because the other slot of it is also occupied, another `relocation` is required. If relocation is a very long chain or meets a infinite loop, the table should be resized or rehashed.

## Schedule

- [x]  Implement the basic `search`, `insert`, `remove` methods.

- [ ]  Implement lock-free `resize`.

- [ ]  Implement other optimizations.

## Reference
* Nguyen, N., & Tsigas, P. (2014). Lock-Free Cuckoo Hashing. 2014 IEEE 34th International Conference on Distributed Computing Systems, 627-636.


rogercloud:
Please move the comment of this PR to the related issue. There might be several PRs related to an issue so it's a better place to describe the problem, the possible solution and the scheules.

francis0407:
This PR is moved to https://github.com/datenlord/lockfree-cuckoohash/pull/1

PR #107, opened 2020-08-30
Add conversion crate, which will be used to fix lint issue

mrwwheat7:


pwang7 (commented):
Is this file copied from sync_fuse?

mrwwheat7:
Changed to use external dependency

pwang7 (approved):

PR #122, opened 2020-10-09
Update aligned-bytes to 0.1.4

Nugine:


async_fuse/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["filesystem"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+aligned-bytes = "0.1.4"
```
pwang7, line 16: how about use git dependency?

Nugine: `aligned-bytes` is an independent public library.
If there is any change in the future, I prefer to bump the version instead of the commit hash.

pwang7 (commented):
fix conflicts please

pwang7 (approved):

PR #160, opened 2020-12-04
Refactor ByteSlice

Nugine:
replace `crate::fuse::byte_slice::ByteSlice` with `crate::fuse::de::Deserializer`


async_fuse/src/fuse/decoder.rs
```diff
@@ -0,0 +1,339 @@
...
+            "no enough bytes to fetch, remaining {} bytes but to fetch {} bytes",
+            len,
+            need,
+        );
+        return Err(DecodeError::NotEnough);
+    }
+    Ok(())
+}
+
+impl<'b> Decoder<'b> {
```
rogercloud, line 80: The functions in the struct Decoder is all about fetch data from u8 slice, while the decode function is actually defined in the T structure. So I suggest to rename this to something related to buf or slice, for example slice generator?

The current method call example is:
```
SomeTypeDeriveDecode a = decoder.decode();
```

I prefer the following one:
```
SomeTypeDeriveDecode a = SomeTypeDeriveDecode::decode(slice_generator);
```
It's more clear that the decode function is defined in the `SomeTypeDeriveDecode`, and the generator only handles the slice stuff.

async_fuse/src/fuse/decoder.rs
```diff
@@ -0,0 +1,339 @@
...
+                trace!("no trailing zero in bytes, cannot fetch c-string");
+                DecodeError::NotEnough
+            })?
+            .wrapping_add(1);
+        debug_assert!(strlen <= self.bytes.len());
+        unsafe { Ok(self.pop_bytes_unchecked(strlen)) }
+    }
+
+    /// Fetch some bytes as a OS-string
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
```
rogercloud, line 205: We can ignore the macos curretnly.

async_fuse/src/fuse/decoder.rs
```diff
@@ -0,0 +1,339 @@
...
+pub trait Decode<'b>: Sized {
+    /// Decode from bytes
+    fn decode(de: &'_ mut Decoder<'b>) -> Result<Self, DecodeError>;
+}
+
+/// The error returned by `Decoder`
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
+pub enum DecodeError {
+    /// Expected more data
+    #[error("NotEnough")]
```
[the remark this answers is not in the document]
Nugine: `DecodeError` is an internal error type but not a report type. (`anyhow::Error` is a report type)
There is more information in the context and logs.

async_fuse/src/fuse/decoder.rs
Nugine: `Decode`/`Decoder` are just like `Deserialize`/`Deserializer` in `serde`.
<https://docs.rs/serde/1.0.118/serde/de/index.html>

byte_slice is not a slice though there is a slice in it.
slice_generator is not a [`Generator`](https://doc.rust-lang.org/stable/std/ops/trait.Generator.html).
I think both of them are misleading.

async_fuse/src/fuse/decoder.rs
Nugine: shall we emit a compile error when compiles on macos?
```rust
// src/lib.rs
#[cfg(target_os = "macos")]
compile_error!("async-fuse does not support macos")
```

async_fuse/src/fuse/decoder.rs
Nugine: How about making `fetch_*` free functions which take `&mut &[u8]` as input?

rogercloud (approved):

Issue #164, opened 2020-12-08
memfs::dir::Dir has no Drop implementation


Nugine:
**Datenlord Version**
master

**Describe the bug and reproduce steps**
memfs::dir::Dir has no Drop implementation, causes memory leak.

**Expected behavior**
Calls closedir(3) when Dir drops

Nugine:
`Dir::from_fd` shoule be unsafe
`Dir` should be Send+Sync

Nugine:
closed by #168

PR #168, opened 2020-12-15
Refactor dir (#164)

Nugine:
refactor dir (#164)
    
* refactor `crate::memfs::dir::Dir`, fix resource leak problem
* change `crate::memfs::fs_util::load_dir_data`, fix use-fd-after-close problem    
* upgrade memchr to 2.3.4
+ add function `with_c_str`, `unblock`, `clear_errno`, `errno` and `nix_to_io_error` in `crate::util`
    
`unblock` is a polyfill from smol 0.1 to 1.2.
We should replace all calls of `smol::blocking!` with `unblock` after this commit.


async_fuse/src/memfs/dir.rs
```diff
@@ -1,63 +1,115 @@
...
+    }
+
+    /// Calls fdopendir(3)
+    /// # Errors
+    /// This method will return `io::Error` if the underlying syscalls fails.
+    /// When an error occurs, closes the `fd` and returns the previous error.
+    /// # Safety
+    /// This function **consumes ownership** of the specified file descriptor.
+    /// The returned object will take responsibility for closing it when the object goes out of scope.
+    pub unsafe fn fdopendir(fd: RawFd) -> io::Result<Self> {
```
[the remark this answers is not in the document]
Nugine: I add `Dir::try_from_raw_fd` in the new commit.

async_fuse/src/memfs/dir.rs
```diff
@@ -1,63 +1,115 @@
...
-    ino: INum,
+    ino: ino_t,
     /// The `SFlag` type of the entry
     entry_type: SFlag,
     /// The entry name
     name: OsString,
 }
 
+/// cast `&[c_char]` to `&[u8]`
+fn cstr_to_bytes(s: &[c_char]) -> &[u8] {
```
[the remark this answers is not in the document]
Nugine: Ok

async_fuse/src/memfs/fs_util.rs
```diff
@@ -271,23 +273,23 @@ pub fn convert_to_fuse_attr(attr: FileAttr) -> FuseAttr {
...
-                _ => None,
-            })
-            .collect();
-        dir_entry_map
-    );
-    Ok(dir_entry_map)
+pub async unsafe fn load_dir_data(fd: RawFd) -> anyhow::Result<BTreeMap<OsString, DirEntry>> {
+    unblock(move || {
+        let dir =
+            Dir::fdopendir(fd).with_context(|| format!("failed to build Dir from fd={}", fd))?;
```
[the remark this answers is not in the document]
Nugine: The new commit uses dup(2). But I'm not sure whether it is safe.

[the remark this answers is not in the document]
Nugine: openat(dirfd, ".", ...) can open a new fd of the same directory

async_fuse/src/memfs/dir.rs
```diff
@@ -1,46 +1,97 @@
...
+unsafe impl Sync for Dir {}
 
-/// Directory meta-data
-#[derive(Debug)]
-pub struct Dir(ptr::NonNull<libc::DIR>);
+impl Drop for Dir {
+    fn drop(&mut self) {
+        let dirp = self.0.as_ptr();
+        unsafe {
+            let _ = libc::closedir(dirp);
```
rogercloud, line 30: We'd better handle the error and log it if necessary.

async_fuse/src/memfs/dir.rs
```diff
@@ -1,46 +1,97 @@
...
+    #[allow(dead_code)]
+    pub fn opendir<P: AsRef<Path>>(path: P) -> io::Result<Self> {
+        let bytes = path.as_ref().as_os_str().as_bytes();
+        with_c_str(bytes, |p| unsafe {
+            let dirname = p.as_ptr();
+            let dirp = libc::opendir(dirname);
+            if dirp.is_null() {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(Self(NonNull::new_unchecked(dirp)))
```
rogercloud, line 48: As the dirp is not null after the preview if check, shall we call new directly instead of new_unchecked?

async_fuse/src/memfs/dir.rs
```diff
@@ -1,46 +1,97 @@
...
+    /// This function **consumes ownership** of the specified file descriptor.
+    /// The returned object will take responsibility for closing it when the object goes out of scope.
+    pub unsafe fn fdopendir(fd: RawFd) -> io::Result<Self> {
+        let dirp = libc::fdopendir(fd);
+        if dirp.is_null() {
+            let err = io::Error::last_os_error();
+            let _ = libc::close(fd);
+            return Err(err);
+        }
+        Ok(Self(NonNull::new_unchecked(dirp)))
```
rogercloud, line 66: Same as above

async_fuse/src/memfs/dir.rs
```diff
@@ -1,46 +1,97 @@
...
+        }
+    }
+}
 
 /// Directory entry
 #[derive(Debug)]
 pub struct DirEntry {
     /// The i-number of the entry
-    ino: INum,
+    ino: ino_t,
```
rogercloud, line 94: This is also a raw data, does it affect the safety of the code?

async_fuse/src/memfs/dir.rs
```diff
@@ -94,58 +153,77 @@ impl DirEntry {
...
+            name: name.to_owned(),
             entry_type,
         }
     }
 }
 
-impl Iterator for Dir {
-    type Item = nix::Result<DirEntry>;
+/// Iterator over the entries in a directory.
+pub struct ReadDir {
```
rogercloud, line 163: Actually it's a Dir Iterator. Let's make the name DirIter.

async_fuse/src/memfs/fs_util.rs
```diff
@@ -271,23 +273,32 @@ pub fn convert_to_fuse_attr(attr: FileAttr) -> FuseAttr {
...
-            .collect();
-        dir_entry_map
-    );
-    Ok(dir_entry_map)
+pub async fn load_dir_data(dirfd: RawFd) -> anyhow::Result<BTreeMap<OsString, DirEntry>> {
+    unblock(move || {
+        let dir = {
+            // new_fd = openat(dirfd, ".", ...)
+            // dirp = fdopendir(new_fd)
+            // it is safe because the new fd is not used by anything other than dirp
```
rogercloud, line 281: Remove it.

async_fuse/src/util.rs
```diff
@@ -72,3 +77,73 @@ pub fn mode_from_kind_and_perm(kind: SFlag, perm: u16) -> u32 {
...
+#[allow(dead_code)]
+pub async fn unblock<T, F>(f: F) -> T
+where
+    F: FnOnce() -> T + Send + 'static,
+    T: Send + 'static,
+{
+    smol::Task::blocking(async move { f() }).await
+}
+
+/// Stores short bytes on stack, stores long bytes on heap and provides [`CStr`]
```
rogercloud, line 91: What's the benefit of storing bytes on stack? It requires another copy from the code.

async_fuse/src/memfs/dir.rs
Nugine: [NonNull::new](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.new) performs a null check which is equivalent to `if-else`.
`if dirp.is_null()` is more explicit.

async_fuse/src/memfs/dir.rs
Nugine: Same as above

async_fuse/src/memfs/dir.rs
Nugine: `RawFd` is a resource handle type while `ino_t` (`INum`/`ino_t`/`u64`) is an index type.
Handles should be managed carefully. But indices have nothing to do with any resource. They are just numbers.

async_fuse/src/memfs/dir.rs
Nugine: The name [`ReadDir`](https://doc.rust-lang.org/std/fs/struct.ReadDir.html)  is from the std library.

<https://rust-lang.github.io/api-guidelines/naming.html#methods-on-collections-that-produce-iterators-follow-iter-iter_mut-into_iter-c-iter>
<https://rust-lang.github.io/api-guidelines/naming.html#iterator-type-names-match-the-methods-that-produce-them-c-iter-ty>

Maybe `IntoIter` is better.

async_fuse/src/memfs/fs_util.rs
Nugine: pingcap style guide explains why here needs some comments.

<https://pingcap.github.io/style-guide/rust/unsafe.html>

> Document every use of unsafe.
> Even trivial uses should have a comment stating why it is safe. Including unsafe impl.
> Document why unsafe blocks are safe and how Rust’s invariants are ensured.
> For unsafe functions, document the invariants the caller must ensure, and any restrictions on the input, or how the output may be used.

async_fuse/src/util.rs
Nugine: Copy is unavoidable when passing bytes to a ffi function which needs c-str, unless the bytes are nul-terminated.
We can reduce allocation by storing short bytes on stack like what `with_nix_path` does.
<https://docs.rs/nix/0.19.1/src/nix/lib.rs.html#234-253>
`with_c_str` returns `io::Result<T>` and will not reject bytes whose length is greater than `PATH_MAX`

async_fuse/src/memfs/dir.rs
Nugine: It an error occurs when dropping `Dir`, what is the expected behavior?
+ ignore?
+ panic?
+ debug_assert?
+ just log and continue?

rogercloud: I recommend panic unless we know how to handle and recover.

Nugine: Panic inside `drop` will cause abort. The whole process will be terminated at once.
We have no chance to leave any information for debugging.

async_fuse/src/util.rs
rogercloud: Pleae tell the benefit of stack allocation. And comment the stack buf size.

rogercloud (approved):

PR #169, opened 2020-12-18
Update CSI smol version to 1.2.5

pzheng1025:


csi/src/main.rs
```diff
@@ -245,7 +245,7 @@ fn run_async_controller_server(controller_server: Server) {
         let f = futures::future::pending::<()>();
         f.await;
     }
-    smol::run(async move {
+    smol::block_on(async move {
```
rogercloud, line 248: Why don't we add compact here?

pzheng1025: We add compact for etcd because etcd_client library depends on tokio.
Don't need for other places without etcd_client.

pzheng1025: And Tian Qi refactor etcd_client to remove dependency of tokio then we don't need compact.

rogercloud (approved):

PR #179, opened 2021-01-12
Unified datenlord error code.

pzheng1025:


csi/src/util.rs
```diff
@@ -215,6 +216,37 @@ pub fn spawn_grpc_task<R: Send + 'static>(
...
 }
 
+/// Send async failure `gRPC` response
+pub async fn async_fail_test<R>(sink: UnarySink<R>, rsc: RpcStatusCode, err: &DatenlordError) {
+    debug_assert_ne!(
+        rsc,
+        RpcStatusCode::OK,
+        "the input RpcStatusCode should not be OK"
+    );
+    let rs = RpcStatus::new(rsc, Some(format!("{}", err)));
```
pzheng1025, line 226: Original err string will contain the context message of anyhow. If we use our error type, the error message will lose context information. Is this OK?

csi/src/worker.rs
```diff
@@ -130,21 +171,12 @@ impl Worker for WorkerImpl {
...
                     .build_volume_from_source(&vol_id_str, vol_name, vol_size, content_source)
-                    .await
-                    .map_err(|(rsc, anyhow_err)| {
-                        let err = anyhow_err.context("failed to create volume from source");
-                        debug!(
-                            "failed to create volume from source, the error is: {}",
-                            util::format_anyhow_error(&err),
-                        );
-                        (rsc, err)
-                    })?;
```
pzheng1025, line 147: We will lost context and error message?

csi/src/worker.rs
```diff
@@ -48,20 +49,11 @@ impl WorkerImplInner {
...
-                    .copy_volume_from_snapshot(vol_size, source_snapshot_id, &vol_id.to_string()).await
-                    .map_err(|(rsc, anyhow_err)| {
-                        (
-                            rsc,
-                            anyhow_err.context(format!(
-                                "failed to populate volume ID={} and name={} from source snapshot ID={}",
-                                vol_id, vol_name, source_snapshot_id,
-                            )),
-                        )
-                    })?;
```
pzheng1025, line 64: Do we need to print the error message if result is Err() ?

csi/src/util.rs
pzheng1025: Or we need add error message to our error code?

csi/src/error.rs
```diff
@@ -0,0 +1,25 @@
...
+//use anyhow::Error;
+use thiserror::Error;
+
+pub type DatenlordResult<T> = Result<T, DatenlordError>;
+#[derive(Error, Debug)]
+pub enum DatenlordError {
+	#[error(transparent)]
+	IoErr(#[from] std::io::Error),
+	#[error("AlreadyExist")]
+	AlreadyExist,
```
Nugine, line 13: What is "AlreadyExist"?

csi/src/error.rs
```diff
@@ -0,0 +1,192 @@
...
+        source: walkdir::Error,
+        context: Option<String>,
+    },
+
+    #[error("AlreadyExist, context is {:?}", .context)]
+    AlreadyExist { context: Option<String> },
+    #[error("Snapshot ID={} not found, context is {:?}", .snapshot_id, .context)]
+    SnapshotNotFound {
+        snapshot_id: String,
+        context: Option<String>,
```
pzheng1025, line 27: Consider to change to Vec<String>.

csi/src/error.rs
```diff
@@ -0,0 +1,290 @@
...
+        source: grpcio::Error,
+        /// Context of the error
+        context: Vec<String>,
+    },
+
+    /// API is not implemented
+    #[error("Not implemented, context is {:?}", .context)]
+    Unimplemented {
+        /// Context of the error
+        context: Vec<String>,
```
rogercloud, line 176: As context is in all enum variant, can we implement it by some macro?

csi/src/error.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+impl DatenLordError {
+    /// Add context for `DatenLordError`
+    pub fn add_context<C>(mut self, ctx: C) -> Self
+    where
+        C: Into<String>,
+    {
+        macro_rules! append_context {
+            ($context: ident, [$($target:ident),*]) => {
```
rogercloud, line 205: this `$context` is not used, right?

csi/src/error.rs
Nugine: Using macro here will break the readability. It's not necessary.

csi/src/error.rs
Nugine: `$context` matches the identifier `ctx` at line 214.

csi/src/error.rs
```diff
@@ -0,0 +1,290 @@
...
+    },
+}
+
+/// Add context to `DatenLordResult`
+pub trait Context<C, T> {
+    /// Add context to `DatenLordResult`
+    fn add_context(self, ctx: C) -> DatenLordResult<T>
+    where
+        C: Into<String>;
+}
```
Nugine, line 186: It's better to provide a lazy evaluated method. See [`anyhow::Context::with_context`](https://docs.rs/anyhow/1.0.38/anyhow/trait.Context.html#tymethod.with_context) and [`clippy::or_fun_call`](https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call).

Without lazy evaluation, we will pay for allocating and formatting lots of error strings even when everything is ok. It can be avoided by a simple method.

Nugine (changes requested):

csi/src/error.rs
pzheng1025: Yes. Agree.

csi/src/error.rs
pzheng1025: Will add.

csi/src/error.rs
pzheng1025: Should change ctx to $context in line 208 to make more sense. Will modify.

Nugine (dismissed):

Nugine (approved):

Nugine (approved):

pwang7 (approved):

PR #189, opened 2021-02-28
Remove anyhow dependency from csi

pzheng1025:


csi/src/bin/bind_mounter.rs
```diff
@@ -1,12 +1,19 @@
 //! The helper command to bind mount for non-root user
 
-use anyhow::{bail, Context};
 use clap::{App, Arg};
 use log::debug;
 use nix::mount::{self, MsFlags};
 use std::ffi::OsStr;
 use std::path::Path;
 
+#[path="../error.rs"]
```
pwang7, line 9: Is there better way to depend on error type?

pzheng1025: Otherwise we need to add src/lib.rs, so that this binary can refer to error module. 
I think this way is better. How do you think?

pwang7 (approved):

PR #192, opened 2021-03-17
Register Node Id and Volume Information in ETCD

rogercloud:
1. Refactor etcd client wrapper and error handling to `common` module.
2. Register Node Id and Volume Information in ETCD.


async_fuse/src/main.rs
```diff
@@ -44,10 +44,146 @@ use log::debug;
...
+/// ETCD ndoe id info
+const ETCD_NODE_ID_INFO_PREFIX: &str = "datenlord_etcd_node_id_info_";
+/// ETCD volume information lock
+const ETCD_VOLUME_INFO_LOCK: &str = "datenlord_etcd_volume_info_lock";
+/// ETCD volume information prefix
+const ETCD_VOLUME_INFO_PREFIX: &str = "datenlord_etcd_volume_info_";
+
+/// Register current node to etcd and get a dedicated node id.
+/// The registered information contains IP.
+async fn register_node_id(etcd_client: &EtcdDelegate) -> anyhow::Result<u64> {
```
pzheng1025, line 73: Is this overlapped with https://github.com/datenlord/datenlord/blob/master/csi/src/meta_data.rs#L139?

pzheng1025:
Later csi and fuse will merge together. We'd better to have a signal place to register information to etcd.

async_fuse/src/main.rs
rogercloud: When will that function be called? And this function also register an IP.

pzheng1025: That is called when the pod is created.

pzheng1025 (approved):

pwang7 (approved):

PR #193, opened 2021-03-17
Add io_uring proactor implementation

Nugine:
This PR introduces a proactor implementation. It is built on the top of [datenlord/ring-io](https://github.com/datenlord/ring-io).

The proactor can execute asynchronous file IO operations, which reduces the overhead of thread pool and syscalls.

- [x] `Proactor` and `IoRequest`
- [x] clippy lints
- [x] documentation

This PR is ready now.


Nugine:
The `proactor` module is designed to be generic and extendable. But now it has not been fully tested and has no benchmark.
In the future, we should move it into [datenlord/ring-io](https://github.com/datenlord/ring-io) or its higher level abstraction .

rogercloud (approved):

PR #203, opened 2021-05-12
Add a cache system

rogercloud:


async_fuse/src/memfs/node.rs
```diff
@@ -434,13 +438,30 @@ impl Node {
...
-        self.need_load_node_data_helper()
+
+        if offset > self.attr.size.cast() {
+            return false;
+        }
+
+        match self.data {
+            NodeData::RegFile(ref cache) => {
+                cache
+                    .get_file_cache(self.name.to_str().unwrap(), offset, len)
```
Nugine, line 455: Unix file name are not always UTF-8 strings.
A GBK file name may cause fatal crashs without messages.

async_fuse/src/test/integration_tests.rs
```diff
@@ -358,8 +366,15 @@ fn test_symlink_dir(mount_dir: &Path) -> anyhow::Result<()> {
     assert_eq!(src_dir, target_path, "symlink target path not match");
 
     let dst_path = Path::new(&dst_dir).join(src_file_name);
-    fs::write(&dst_path, FILE_CONTENT)
-        .context(format!("failed to write to file={:?}", src_path))?;
+    match fs::write(&dst_path, FILE_CONTENT)
+        .context(format!("failed to write to file={:?}", dst_path))
+    {
+        Err(err) => {
+            std::thread::sleep(std::time::Duration::from_secs(10));
```
Nugine, line 373: Why?

async_fuse/src/fuse/fuse_reply.rs
```diff
@@ -22,118 +20,154 @@ use super::protocol::{
...
             };
+
             let h = utilities::cast_to_ptr(&header);
             let header_bytes = unsafe { slice::from_raw_parts(h, header_len) };
-            let iovecs: Vec<_> = if data_len > 0 {
-                vec![IoVec::from_slice(header_bytes), IoVec::from_slice(bytes)]
+            let iovecs: Vec<_> = if data.len() > 0 {
+                let mut header = vec![IoVec::from_slice(header_bytes)];
+                header.append(&mut data.as_io_vec_list());
+                header
```
Nugine, line 137: We can insert the header at index 0 to reduce allocation

async_fuse/src/fuse/fuse_reply.rs
```diff
@@ -229,64 +263,105 @@ impl_fuse_reply_error_for! {
...
+            #[allow(dead_code)]
+            fn as_io_vec(&self) -> IoVec<&[u8]> {
+                let type_len = mem::size_of::<$t>();
+                let bytes: &[u8] = match type_len {
+                    0 => &[],
+                    l => {
+                        let p = utilities::cast_to_ptr(self);
+                        unsafe { slice::from_raw_parts(p, l) }
+                    }
+                };
```
Nugine, line 278: There is a safe wrapper `crate::fuse::abi_marker::as_abi_bytes`.

https://github.com/datenlord/datenlord/blob/28be4f773d78ed3116b8c63f4df88e26fa7f57fc/async_fuse/src/fuse/abi_marker.rs#L25-L30

async_fuse/src/memfs/fs_util.rs
```diff
@@ -289,20 +289,20 @@ pub async fn load_dir_data(dirfd: RawFd) -> anyhow::Result<BTreeMap<OsString, Di
...
 
 /// Helper function to load file data
-pub async fn load_file_data(fd: RawFd, file_size: usize) -> anyhow::Result<Vec<u8>> {
+pub async fn load_file_data(fd: RawFd, offset: usize, len: usize) -> anyhow::Result<Vec<u8>> {
     let file_data_vec = smol::unblock(move || {
-        let mut file_data_vec: Vec<u8> = Vec::with_capacity(file_size);
+        let mut file_data_vec: Vec<u8> = Vec::with_capacity(len);
         unsafe {
             file_data_vec.set_len(file_data_vec.capacity());
         }
```
Nugine, line 297: Potential unsoundness here.

See <https://doc.rust-lang.org/std/vec/struct.Vec.html#safety-2>

async_fuse/src/main.rs
```diff
@@ -249,20 +266,23 @@ fn main() -> anyhow::Result<()> {
...
         Some(vi) => vi,
         None => panic!("No volume information input"),
     };
-
     let etcd_delegate = EtcdDelegate::new(etcd_address_vec)?;
-
     debug!("FUSE mount point: {}", mount_point);
 
+    let cache_capacity = match matches.value_of(CACHE_CAPACITY_ARG_NAME) {
+        Some(vi) => vi.parse::<usize>().unwrap(),
```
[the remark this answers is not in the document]
rogercloud: We can panic or provide a default value on any error. Here I choose to panic.

async_fuse/src/test/integration_tests.rs
rogercloud: It's used to debug, let me delete it.

async_fuse/src/memfs/node.rs
rogercloud: Yes, that's true, but it almost impossible to know the encoding, actually OS treat it as a sequence of bytes. After all it's a good point, a possible improvement is to adjust the key of the cache to a byte array.

async_fuse/src/fuse/fuse_reply.rs
rogercloud: Good advice.

async_fuse/src/fuse/fuse_reply.rs
rogercloud: Yes, it works for all FuseAbiData, but there's a tuple in this case, `(FuseEntryOut, FuseOpenOut)`, which is not covered.

async_fuse/src/main.rs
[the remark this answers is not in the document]
Nugine: It's better to exit with an error message.
<https://docs.rs/clap/2.33.3/clap/macro.value_t.html>

Derive macros can reduce most of the lines used to setup clap.
<https://docs.rs/structopt/0.3.21/structopt/>
<https://github.com/clap-rs/clap#using-derive-macros>

async_fuse/src/memfs/fs_util.rs
rogercloud: Yes, I guess you mentioned it before. Let me set the length after loading the file, just like what they do in the example.

async_fuse/src/fuse/fuse_reply.rs
Nugine: Tuples do not have any guarantees about their layout. Transmuting them is **UB**.
<https://doc.rust-lang.org/reference/type-layout.html#tuple-layout>
It is impossible to impl `AsIoVec` for `(FuseEntryOut, FuseOpenOut)`.

async_fuse/src/memfs/fs_util.rs
Nugine: The simplest answer is `vec![0; len]`.

async_fuse/src/fuse/fuse_reply.rs
rogercloud: Let me add this tuple to the trait FuseAbiData implementation.

async_fuse/src/memfs/node.rs
Nugine: Another choice is forbidding non-UTF8 file names. I'm not sure which is better.

async_fuse/src/fuse/fuse_reply.rs
Nugine: No. `(FuseEntryOut, FuseOpenOut)` does not satisfy the safety requirements of `FuseAbiData`.

You may need to write a struct like this:

```rust
#[repr(C)] // layout guarantee
pub struct Pair<T1, T2>{
    first: T1,
    second: T2
}

impl<T1,T2> Pair<T1,T2>{
   pub fn new(first: T1, second: T2) -> Self {
       assert_eq!(align_of::<T1>() == align_of::<T2>()); // Ensure there are no padding bytes.
       ...
   }
}

impl AsIoVec for Pair<FuseEntryOut, FuseOpenOut> {
   ...
}
```

async_fuse/src/main.rs
```diff
@@ -204,6 +209,18 @@ fn main() -> anyhow::Result<()> {
...
         )
+        .arg(
+            clap::Arg::with_name(CACHE_CAPACITY_ARG_NAME)
+                .short("c")
+                .long(CACHE_CAPACITY_ARG_NAME)
+                .value_name("CACHE_CAPACITY")
+                .takes_value(true)
+                .required(false)
+                .help(
+                    "Set the mount point of FUSE, \
```
mrwwheat7, line 220: Looks like this is for Cache capacity, right?

async_fuse/src/fuse/fuse_reply.rs
```diff
@@ -22,118 +20,154 @@ use super::protocol::{
...
             let header_bytes = unsafe { slice::from_raw_parts(h, header_len) };
-            let iovecs: Vec<_> = if data_len > 0 {
-                vec![IoVec::from_slice(header_bytes), IoVec::from_slice(bytes)]
+            let iovecs: Vec<_> = if data.len() > 0 {
+                let mut header = vec![IoVec::from_slice(header_bytes)];
+                header.append(&mut data.as_io_vec_list());
+                header
             } else {
                 vec![IoVec::from_slice(header_bytes)]
             };
```
Nugine, line 140: The latter allocation can be reduced by using on-stack DST coercion.
Here is an example:
```rust
let cond = false;

let (v1, v2);
let s: &[u8] = if cond {
    v1 = vec![1, 2];
    &*v1
} else {
    v2 = [3];
    &v2
};

println!("{:?}", s);
```

async_fuse/src/main.rs
rogercloud: Yes, I fixed it in later commit, not included in this PR. Let me fix it here. Thanks.

async_fuse/src/memfs/node.rs
rogercloud: It's defined by the users not even OS. OS only accepts bytes, so sticking to the bytes is a better choise from my view.

async_fuse/src/memfs/fs_util.rs
rogercloud: The initialized is not necessary.

async_fuse/src/fuse/fuse_reply.rs
```diff
@@ -7,133 +7,186 @@ use nix::sys::uio::{self, IoVec};
...
+
+            let header = FuseOutHeader {
+                len: (header_len).cast(),
+                error: crate::util::convert_nix_errno_to_cint(error_code).overflow_neg(), // FUSE requires the error number to be negative
+                unique: self.unique,
+            };
+
+            let h = utilities::cast_to_ptr(&header);
+            let header_bytes = unsafe { slice::from_raw_parts(h, header_len) };
+            let iovecs: Vec<_> = vec![IoVec::from_slice(header_bytes)];
```
Nugine, line 183: unnecessary allocation

Nugine (changes requested):
Format and lint checks failed.
And FUSE test is hanging for 3 hours by now. What is going wrong?

async_fuse/src/fuse/fuse_reply.rs
Nugine: Line 148, 149, 181 and 182 can be replaced by `as_abi_bytes`

async_fuse/Cargo.toml
```diff
@@ -36,6 +36,8 @@ smol = { version = "1.2.5" }
 thiserror = "1.0.22"
 utilities = { git = "https://github.com/datenlord/utilities", rev = "4ef408d" }
 common = {path = "../common"}
+lockfree-cuckoohash = {git = "https://github.com/datenlord/lockfree-cuckoohash", rev = "e5849f18"}
+priority-queue = "1.1.0"
```
pzheng1025, line 40: put it in alphabetic order

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+const MEMORY_BUCKET_SIZE_IN_BYTE: usize = MEMORY_BUCKET_VEC_SIZE * MEMORY_BLOCK_SIZE_IN_KB * 1024;
+/// The default capacity in bytes, 10GB
+const GLOBAL_CACHE_DEFAULT_CAPACITY: usize = 10 * 1024 * 1024 * 1024;
+
+/// A mapping from file name to per-file memory block mapping
+pub struct GlobalCache {
+    /// Map from file identifier to cache content map
+    inner: HashMap<Vec<u8>, HashMap<usize, MemBlockBucket>>,
+    /// Priority queue to track bucket usage
+    queue: Mutex<PriorityQueue<MemBlockBucket, i64>>,
```
pzheng1025, line 37: Do we need to use async mutex here?

rogercloud: I'll modify the lock to async lock in latter commit. This commit just prooves the cache concept.

async_fuse/Cargo.toml
rogercloud: Good point! Guess there's no lint for it.

Nugine (approved):

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,742 @@
...
+        }
+
+        if let Some(ref block) = self.inner {
+            let ptr = block.as_ptr_from_offset(self.offset);
+            unsafe { std::slice::from_raw_parts(ptr, self.end.overflow_sub(self.offset)) }
+        } else {
+            panic!("the inner Option<MemBlock> should not be none");
+        }
+    }
+}
```
Nugine, line 497: **Internal unsoundness**: No matter where it is used, this method should be `unsafe` because it bypasses the `RwLock`.

For this case, `with`-style methods are better than `unsafe` methods.
https://doc.rust-lang.org/nightly/std/thread/struct.LocalKey.html#method.with

Nugine (changes requested):

scripts/datenlord.yaml
```diff
@@ -221,6 +221,8 @@ spec:
               value: csi-etcd-svc:2379
             - name: FUSE_VOLUME_INFO
               value: https://mybucket
+            - name: CACHE_CAPACITY
+              value: "10737418240"
```
pzheng1025, line 225: Add a comment for this value.

async_fuse/src/memfs/cache/mod.rs
rogercloud: It's a good point, we should expose unsafe attribute to the caller.

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+                    .for_each(&mut copy_fn);
+            };
+        }
+
+        self.size
+            .fetch_add(grow_memory.overflow_mul(self.block_size), Ordering::Relaxed);
+
+        let mut dealloc_fn = |b: &mut Option<MemBlock>| {
+            if b.is_some() {
+                b.take();
```
mrwwheat7, line 341: Maybe a dumb question, right here we just delete the block in the bucket, right? Then the bucket still exists, and if we call `get_file_cache` function now, this empty bucket may still be returned?

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+
+                        let e = if k == end_key
+                            && pos == end_index.overflow_sub(start_index).overflow_sub(1)
+                        {
+                            end_offset.overflow_add(1)
+                        } else {
+                            self.block_size
+                        };
+
+                        result.push(IoMemBlock::new(block, s, e));
```
mrwwheat7, line 186: Just want to check if my understanding is correct, this block can also be None, right? Since it may be deleted because of Cache capability.

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+
+impl IoMemBlock {
+    /// The constructor of `IoMemBlock`
+    const fn new(inner: Option<MemBlock>, offset: usize, end: usize) -> Self {
+        Self { inner, offset, end }
+    }
+
+    /// Turn `IoMemBlock` into slice
+    #[allow(dead_code)]
+    pub(crate) unsafe fn as_slice(&self) -> &[u8] {
```
mrwwheat7, line 485: Will this function be used on the return value of `get_file_cache` function? Since looks like the return value of `get_file_cache` function may be None, then this function will panic, is it Ok?

async_fuse/src/memfs/node.rs
```diff
@@ -1050,12 +1094,12 @@ impl Node {
     // File only methods
 
     /// Get file data
-    pub fn get_file_data(&self) -> &Vec<u8> {
+    pub fn get_file_data(&self, offset: usize, len: usize) -> Vec<IoMemBlock> {
```
mrwwheat7, line 1097: A little confused here, some element in the `get_file_cache` return value could be None, right? Means there's no buffer of this range in the Cache. If so, this `get_file_data` function may only return part of the file data, is it expected?

Nugine (approved):

scripts/datenlord.yaml
```diff
@@ -221,6 +221,9 @@ spec:
               value: csi-etcd-svc:2379
             - name: FUSE_VOLUME_INFO
               value: https://mybucket
+              # The copacity for the cache in memfs
```
pzheng1025, line 224: typo capacity. Actually I mean to provide a comment for the long number " 10737418240" like "request 10G cache"

async_fuse/src/memfs/node.rs
mrwwheat7: Just after we call the `load_data`, then the data in Cache is deleted because of Cache capability, then we call this `get_file_data`, shouldn't we read the data from file, update Cache and return?

async_fuse/src/memfs/cache/mod.rs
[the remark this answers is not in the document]
rogercloud: Your understanding is right. The memblock is deleted from the bucket, any `get` operation after `delete` will get None in the bucket.

async_fuse/src/memfs/cache/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+                    bucket_lock
+                        .iter_mut()
+                        .filter(|b| b.is_some())
+                        .for_each(&mut dealloc_fn);
+                    exceed = exceed
+                        .overflow_sub(removed_count.overflow_mul(self.block_size).cast::<i64>());
+                }
+                None => break,
+            };
+        }
```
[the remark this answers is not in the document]
rogercloud: Nice catch, I forget to check the size in the test. Thanks.

async_fuse/src/memfs/cache/mod.rs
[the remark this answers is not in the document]
rogercloud: And no bucket will be returned, only memory blocks are returned.

async_fuse/src/memfs/cache/mod.rs
rogercloud: Yes, but we really check the data before the usage, it's guaranteed by the users, such as memfs.

async_fuse/src/memfs/cache/mod.rs
rogercloud: Like the explaination before, if there's None, the system will panic, but we'll check before any usage.

async_fuse/src/memfs/node.rs
rogercloud: Any data loaded in the `load_data` should be in the front end of the priority queue, which should not be deleted. 

We may load partial data from the file instead of the whole requests for the user, in case that the cache has holded part of the requests range. So we can't return the data from file directly. In short the operation is separated into two steps:

1. fill the cache.
2. read data in the cache.

mrwwheat7: I see, I mean what if we first call `load_data`, the data is loaded in Cache, but before we call `get_file_data`, the data in Cache is evicted (Maybe because of other write operations), then right now we call `get_file_data`, wouldn't it be a problem? (some of the data is lost)

pzheng1025 (approved):

Nugine (approved):

PR #207, opened 2021-05-31
Add entrypoint script

pzheng1025:
command line arguments can be provided by datenlord.yaml

### DIFF
--- a/scripts/datenlord-entrypoint.sh
+++ b/scripts/datenlord-entrypoint.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+#set -xv # enable debug
+#set -e # exit on error
+#set -u # unset var as error
+
+ARGS=$@
+while [ $# -gt 0 ]; do
+    key=$1
+    case $key in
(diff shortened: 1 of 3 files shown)


scripts/datenlord-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+#set -xv # enable debug
+#set -e # exit on error
+#set -u # unset var as error
+
+ARGS=$@
+while [ $# -gt 0 ]; do
+	key=$1
```
pwang7, line 9: do not use tab, use four space instead

pwang7, line 9: replace tab by 4 spaces

pwang7 (changes requested):

scripts/datenlord-entrypoint.sh
pzheng1025: Done

pwang7 (approved):

PR #209, opened 2021-06-07
Deploy etcd before DatenLord.

pzheng1025:
DatenLord rely on etcd. Deploy etcd before DatenLord
to avoid DatenLord crash.


pwang7 (changes requested):
How about change this to a shell script?

pzheng1025:
This has been implemented in another PR.

PR #213, opened 2021-06-19
Increase cron stability

pzheng1025:
Reduce etcd replica to 1.
Remove e2e parallel test.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -240,6 +240,9 @@ jobs:
           kubectl apply -f datenlord-setup-snapshot-controller.yaml
           kubectl wait --for=condition=Ready pod -l app=snapshot-controller -n kube-system --timeout=60s
           kubectl get pods -A -o wide
+          # Sleep 60 to wait cluster become stable
+          sleep 60
+          kubectl get pods -A -o wide
           # sudo netstat -lntp && ls -lsh
       - name: CSI E2E Test
         run: |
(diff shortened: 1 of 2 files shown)


rogercloud (dismissed):

.github/workflows/cron.yml
```diff
@@ -247,7 +247,7 @@ jobs:
           tar zxvf kubernetes-test-linux-amd64.tar.gz
           kubectl config view --raw > $K8S_CONFIG
           #kubernetes/test/bin/e2e.test -v=5 -ginkgo.failFast -ginkgo.failOnPending -ginkgo.debug -ginkgo.v -ginkgo.focus='External.Storage' -kubectl-path=`which kubectl` -kubeconfig=$K8S_CONFIG -storage.testdriver=`realpath $E2E_TEST_CONFIG`
-          kubernetes/test/bin/ginkgo -p -v -failFast -failOnPending -debug -focus='External.Storage' -skip='.*Dynamic PV.*filesystem.*should access to two volumes with the same volume mode and retain data across pod recreation on .* node' kubernetes/test/bin/e2e.test -- -v=5 -kubectl-path=`which kubectl` -kubeconfig=`realpath $K8S_CONFIG` -storage.testdriver=`realpath $E2E_TEST_CONFIG` -test.parallel=3
+          kubernetes/test/bin/ginkgo -p -v -failFast -failOnPending -debug -focus='External.Storage' -skip='.*Dynamic PV.*filesystem.*should access to two volumes with the same volume mode and retain data across pod recreation on .* node' kubernetes/test/bin/e2e.test -- -v=5 -kubectl-path=`which kubectl` -kubeconfig=`realpath $K8S_CONFIG` -storage.testdriver=`realpath $E2E_TEST_CONFIG`
```
pwang7, line 250: why remove test.parallel?

pzheng1025: I guess test.parallel causes some stability issues because some tests cannot run in parallel. 
I also test locally w/ and w/o test.parallel. Without test.parallel is more stable and test time doesn't have difference.

pzheng1025: -test.parallel=3 is to set the number of parallel and -p is to enable parallel. It is better to let ginkgo choose the number base on system.

rogercloud (approved):

pwang7 (approved):

PR #224, opened 2021-08-26
Add datenlord metrics exporter

mrwwheat7:


async_fuse/src/main.rs
```diff
@@ -284,6 +293,17 @@ fn main() -> anyhow::Result<()> {
         None => CACHE_DEFAULT_CAPACITY,
     };
 
+    smol::spawn(Compat::new(async move {
```
rogercloud, line 296: Please move this part to the metrics internal so that the metrics service only takes one line in the main function.

rogercloud (changes requested):
So your plan is to add the tests in the next commit?

async_fuse/src/main.rs
mrwwheat7: DONE.

mrwwheat7:
> So your plan is to add the tests in the next commit?

Yes, the test is in the next PR

async_fuse/src/main.rs
```diff
@@ -193,6 +196,7 @@ async fn register_volume(
     Ok(())
 }
 
+#[allow(clippy::too_many_lines)] //allow this for argument parser function as there is no other logic in this function
```
rogercloud, line 199: can be removed

rogercloud (approved):

async_fuse/src/main.rs
mrwwheat7: Done.

rogercloud (approved):

PR #240, opened 2021-10-18
upgrade grpcio and build on latest stable rustc version

devillove084:
rustc 1.55.0 (c8dfcfe04 2021-09-06)
grpcio 0.9.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,11 +25,11 @@ crossbeam-channel = "0.5.0"
 crossbeam-queue = "0.3.1"
 crossbeam-utils = "0.8.1"
 env_logger = "0.8.1"
-etcd-client = { git="https://github.com/datenlord/etcd-client", rev = "c18bfb0"}
+etcd-client = { git="https://github.com/datenlord/etcd-client"}
 event-listener = "2.5.1"
 flate2 = "1.0.16"
 futures = "0.3.5"
-grpcio = { version = "0.8.0", default-features = false, features = ["protobuf-codec"] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -63,7 +63,7 @@ walkdir = "2.3.1"
 protoc-grpcio = "3.0.0"
 
 [dev-dependencies]
-mock-etcd = { git = "https://github.com/datenlord/etcd-client", rev = "c18bfb0" }
+mock-etcd = { git = "https://github.com/datenlord/etcd-client"}
```
pzheng1025, line 66: Don't remove revision

src/main.rs
```diff
@@ -373,10 +373,10 @@ fn get_args(matches: &ArgMatches) -> CliArgs {
 
     let cache_capacity = match matches.value_of(CACHE_CAPACITY_ARG_NAME) {
         Some(cc) => cc.parse::<usize>().unwrap_or_else(|_| {
-            panic!(format!(
```
pzheng1025, line 376: The same.

src/csi/node.rs
```diff
@@ -94,10 +94,10 @@ impl NodeImplInner {
         let add_ephemeral_res = self.meta_data.add_volume_meta_data(vol_id, &volume).await;
         debug_assert!(
             add_ephemeral_res.is_ok(),
-            format!(
+
```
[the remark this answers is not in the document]
devillove084: rust version to 1.47 is also support this feature

Cargo.toml
devillove084: So must merge etcd-client, then i can check revision.

PR #250, opened 2021-12-07
Fix remove inode remotely issue.

pzheng1025:


src/async_fuse/memfs/metadata.rs
```diff
@@ -753,13 +784,18 @@ impl MetaData for DefaultMetaData {
...
+        let (_in_trash, inode) = if let Some(inode) = cache.get_mut(&ino) {
+            (false, inode)
+        } else if let Some(inode) = trash.get_mut(&ino) {
+            (true, inode)
+        } else {
             panic!(
-                "write() found fs is inconsistent, \
-                    the i-node of ino={} should be in cache",
-                ino,
+                "write() found fs is inconsistne, \
```
pzheng1025, line 794: typo

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -731,6 +739,60 @@ impl ReplyXAttr {
...
+#[cfg(feature = "abi-7-18")]
+impl FuseDeleteNotification {
+    /// Create `FuseDeleteNotification`
+    #[must_use]
+    pub const fn new(fd: RawFd) -> Self {
+        Self { fd }
+    }
+
+    /// Notify kernel
+    pub async fn notify(&self, parent: u64, child: u64, name: String) -> nix::Result<usize> {
```
rogercloud, line 759: Can we reuse the raw reply? Once the message implement AsIoVecList trait it can be send via reply method.

src/async_fuse/memfs/node.rs
```diff
@@ -209,10 +215,11 @@ impl DefaultNode {
             attr,
             data,
             fd,
-            // lookup count set to 1 by creation
-            open_count: AtomicI64::new(1),
+            // open count set to 0 by creation
+            open_count: AtomicI64::new(0),
             // open count set to 1 by creation
             lookup_count: AtomicI64::new(1),
+            deferred_deletion: AtomicBool::new(false),
```
rogercloud, line 222: If all the operation on this field is protected by a lock, we can use bool directly

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -554,96 +571,100 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
...
+                inode.get_open_count(),
+                inode.get_lookup_count(),
             );
+            inode.mark_deferred_deletion();
+            // Notify kernel to drop cache
+            if from_remote && inode.get_lookup_count() > 0 {
+                let fuse_fd = *self.fuse_fd.lock().await;
+                // fuse_fd must be set
+                assert!(fuse_fd > 0_i32);
+                #[cfg(feature = "abi-7-18")]
```
rogercloud, line 651: We can wrap two lines into a block and write cfg once.

src/async_fuse/memfs/s3_node.rs
```diff
@@ -21,15 +21,11 @@ use nix::sys::stat::SFlag;
 use smol::lock::RwLock;
 use std::collections::BTreeMap;
 use std::collections::VecDeque;
+use std::os::unix::io::RawFd;
```
rogercloud, line 24: Why do we reformat the use in this matter?

src/async_fuse/memfs/metadata.rs
```diff
@@ -820,87 +845,83 @@ impl DefaultMetaData {
 
     /// Helper function to delete or deferred delete node
     async fn may_deferred_delete_node_helper(&self, ino: INum) -> anyhow::Result<()> {
-        let parent_ino: INum;
-        let node_name: String;
-        let mut deferred_deletion = false;
-        {
+        // TODO: support thread-safe to avoid race condition
+        let (deferred_deletion, parent_ino, node_name) = {
```
rogercloud, line 849: Can we abstract this block into a function? It's too long

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -134,30 +136,44 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
         &self.cache
     }
 
+    /// Set fuse fd into `MetaData`
+    async fn set_fuse_fd(&self, fuse_fd: RawFd) {
+        *self.fuse_fd.lock().await = fuse_fd;
+    }
+
     /// Delete node from trash
     async fn delete_trash(&self, ino: INum) -> bool {
```
rogercloud, line 145: This function seems an attemption, shall we rename it to "try_delete_trash"?

rogercloud: And please double check it's called everywhere that "lookup count" or "open count" is decreased including the local and remote triggers.

src/async_fuse/fuse/fuse_reply.rs
pzheng1025: notification use error as notification as notification type. This is a little different from raw reply. 
I can add a parameter to raw reply to allow pass-in error code.

src/async_fuse/memfs/s3_metadata.rs
pzheng1025: Yes. everywhere has been checked. Will change to try_delete_trash. I think try_delete_node is better because we don't have trash anymore.

src/async_fuse/memfs/s3_node.rs
pzheng1025: I don't have strong opinion on this. I think it is more clearer. If we use old style. We need compress  all std::{} together. I have never seen this style.

src/async_fuse/memfs/node.rs
pzheng1025: Yes. It is possible. But I want to keep for two reasons. 1. make API as &self instead of &mut self. 2. it is possible to get read lock instead of write log to mark deferred_deletion.

src/async_fuse/memfs/s3_metadata.rs
pzheng1025: Yes.

src/async_fuse/memfs/metadata.rs
pzheng1025: Done

src/async_fuse/memfs/s3_metadata.rs
pzheng1025: Done

src/async_fuse/fuse/fuse_reply.rs
pzheng1025: Done

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -160,59 +164,51 @@ impl ReplyRaw {
...
-            let (single, mut vecs);
-            let iovecs: &[IoVec<&[u8]>] = if data.len() > 0 {
-                vecs = data.as_io_vec_list();
-                vecs.insert(0, IoVec::from_slice(header_bytes));
-                vecs.as_slice()
-            } else {
-                single = [IoVec::from_slice(header_bytes)];
-                &single
-            };
+            let mut iovecs = vec![IoVec::from_slice(header_bytes)];
```
rogercloud, line 185: This operation will copy the whole slice, do we really need to copy it?

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -160,59 +164,51 @@ impl ReplyRaw {
...
 
             let header_bytes = abi_marker::as_abi_bytes(&header);
 
-            let (single, mut vecs);
-            let iovecs: &[IoVec<&[u8]>] = if data.len() > 0 {
-                vecs = data.as_io_vec_list();
-                vecs.insert(0, IoVec::from_slice(header_bytes));
-                vecs.as_slice()
-            } else {
-                single = [IoVec::from_slice(header_bytes)];
```
rogercloud, line 183: In this case, we don't copy any data.

src/async_fuse/memfs/node.rs
rogercloud: Good point

src/async_fuse/memfs/s3_node.rs
rogercloud: I prefer the previous version as it requires less typing and the organization is more clear, but I still didn't get my vscode work it out automatically.

src/async_fuse/fuse/fuse_reply.rs
pzheng1025: Done

src/async_fuse/fuse/fuse_reply.rs
pzheng1025: Done

rogercloud (approved):

rogercloud (approved):

pwang7 (approved):

PR #252, opened 2021-12-19
Fix several thread-safe issues.

pzheng1025:
Get one lock and do all things before drop.

### DIFF
--- a/src/async_fuse/memfs/s3_node.rs
+++ b/src/async_fuse/memfs/s3_node.rs
@@ -18,7 +18,7 @@ use log::{debug, warn};
 use nix::fcntl::OFlag;
 use nix::sys::stat::Mode;
 use nix::sys::stat::SFlag;
-use smol::lock::RwLock;
+use smol::lock::RwLockWriteGuard;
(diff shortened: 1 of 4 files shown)


src/async_fuse/memfs/s3_node.rs
```diff
@@ -377,72 +377,66 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3Node<S> {
 pub async fn rename_fullpath_recursive<S: S3BackEnd + Send + Sync + 'static>(
     ino: INum,
     parent: INum,
-    cache: &RwLock<BTreeMap<INum, S3Node<S>>>,
+    cache: &mut RwLockWriteGuard<'_, BTreeMap<u64, S3Node<S>>>,
```
rogercloud, line 380: Why do we use u64 instead of INum here?

pzheng1025: Should be INum.

rogercloud (approved):

pwang7 (approved):

PR #355, opened 2023-03-15
Changed FileAttr to shared Arc<parking_lot::RwLock<...>>

ActivePeter:
## effect
- parent dir and child file node can share a same reference and have agreement on it
- parent dir contains all children‘s file attr to accelerate the process of get child's attr , and we don't need to load attr from remote because each time lookup the parent dir with attrs is already in cache.

### DIFF
--- a/src/async_fuse/memfs/metadata.rs
+++ b/src/async_fuse/memfs/metadata.rs
@@ -1,7 +1,7 @@
 use super::cache::GlobalCache;
 use super::dir::DirEntry;
 use super::dist::server::CacheServer;
-use super::fs_util;
+use super::fs_util::{self, FileAttr};
 use super::node::{self, DefaultNode, Node};
 use super::RenameParam;
 use crate::async_fuse::fuse::protocol::{FuseAttr, INum, FUSE_ROOT_ID};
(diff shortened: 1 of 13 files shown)


src/async_fuse/memfs/metadata.rs
```diff
@@ -15,6 +15,7 @@ use nix::errno::Errno;
 use nix::fcntl::OFlag;
 use nix::sys::stat::SFlag;
 use nix::unistd;
+use parking_lot::RwLock as SyncRwLock;
```
rogercloud, line 18: Why do we rename it?

src/async_fuse/memfs/s3_node.rs
```diff
@@ -27,8 +28,8 @@ use std::sync::Arc;
 use std::time::SystemTime;
 use tokio::sync::RwLockWriteGuard;
 
-/// Block size constant
-const BLOCK_SIZE: usize = 1024;
+// /// Block size constant
```
rogercloud, line 31: Please remove it if it's not required.

rogercloud (commented):
1. What's the paTools submodule?
2. Please rebase and refine the commits before merge.

src/async_fuse/memfs/metadata.rs
ActivePeter: Conflict with tokio RwLock

PR #363, opened 2023-04-20
Refactor : Manage metadata using a distributed K/V engine

xiaguan:
In order to ensure the fault tolerance and consistency of metadata, we plan to use distributed K/V engine to reconstruct metadata.

### DIFF
--- a/src/async_fuse/memfs/kv_engine/mod.rs
+++ b/src/async_fuse/memfs/kv_engine/mod.rs
@@ -4,49 +4,17 @@ use core::fmt::Debug;
 use datenlord::common::error::{Context, DatenLordResult};
 use etcd_client::TxnCmp;
 use serde::{Deserialize, Serialize};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::fmt;
-use std::path::PathBuf;
+
 use std::sync::Arc;
 
-use super::serial::*;
+use super::serial::{SerialDirEntry, SerialFileAttr, SerialNode};
 
-/// In order to derive Serialize and Deserialize,
-/// Replace the 'BTreeMap<String, DirEntry>' with 'HashMap<String, RawDirEntry>'
-#[derive(Serialize, Deserialize, Debug)]
-pub enum RawS3NodeData {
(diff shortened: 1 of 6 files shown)


ActivePeter:
Link the design issue

xiaguan:
#369

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                TxnCmp::Equal,
+                version,
+            );
+        }
+        // add the write operations
+        for (key, value) in &self.buffer {
+            let key = key.clone();
+            if let Some(ref value) = *value {
+                // TODO(xiaguan) : maybe too many clone?
+                let put_req = etcd_client::EtcdPutRequest::new(key.clone(), value.clone());
```
rogercloud, line 167: We can consider reference if it's not shared during this operation.

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        let req = etcd_client::EtcdGetRequest::new(key.clone());
+        let mut resp = self
+            .client
+            .kv()
+            .get(req)
+            .await
+            .with_context(|| format!("failed to get GetResponse from etcd, key={key:?}"))?;
+        let kvs = resp.take_kvs();
+        // panic here because we don't expect to have multiple values for one key
+        assert!(kvs.len() <= 1, "multiple values for one key");
```
rogercloud, line 124: Return an error for bad response?

xiaguan: I added `assert` statements for the sake of convenience during development and debugging.They will not be effective in the release versionThese assert statements describe certain scenarios that are not supported by kv_engine, such as `path2inum(key)`. In these scenarios, the caller expects to receive either none or one value, but not two values. Even if the caller were to handle this situation as an error, they would not be able to fix it, as it is considered undefined behavior.

xiaguan: `panic` is comment typo error . I will fix it .

rogercloud (approved):

PR #370, opened 2023-05-05
Refactor : removing the concurrency safety guarantee of `Node`

xiaguan:
Removing the concurrency safety guarantee of `Node`
Although the changes made in this pull request can pass compilation, we cannot guarantee the correctness of the logic, including passing the tests. The reason for this is that the changes involve a significant amount of code modifications in the kv_engine meta layer. To facilitate the code review process, we have split the changes into smaller pull requests for submission. This approach will enable each pull request's changes to be more easily reviewed and tested, ensuring the overall code quality


xiaguan:
#371

PR #374, opened 2023-05-12
Refactor DatenLord Error Handling with DatenLordError

xiaguan:
# Description:
This pull request refactors the error handling of the DatenLord to align with the general error handling approach of the project.
Previously, the MetaData trait was returning results of type anyhow::Result, while other parts of the codebase were returning DatenLordResult, leading to an interface mismatch and increased complexity. Additionally, using anyhow::Result provided no error type information, making it difficult to discern the exact nature of the error.
We've added a new error type, InternalErr, under DatenLordError. This new error type contains the source anyhow::Error and its context Vec<String>, effectively encapsulating the original anyhow::Error.
# Changes:
- Refactored  functions that were returning anyhow::Result to now return DatenLordResult with the appropriate error type.
- Added InternalErr under DatenLordError to encapsulate anyhow::Error and its context.
- Updated all relevant error handling and propagation logic to match the new error handling approach.
Please review these changes and provide any feedback you may have. We're happy to make further adjustments as necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,14 +25,18 @@ crossbeam-channel = "0.5.0"
 crossbeam-queue = "0.3.1"
 crossbeam-utils = "0.8.1"
 env_logger = "0.10.0"
-etcd-client = { git="https://github.com/datenlord/etcd-client", rev = "f697899"}
+etcd-client = { git = "https://github.com/datenlord/etcd-client", rev = "f697899" }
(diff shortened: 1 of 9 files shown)


ActivePeter:
Seems some tests didn't runs ok.

rogercloud (approved):

rogercloud (approved):

PR #375, opened 2023-05-15
Using distributed KV to rebuild metadata.

xiaguan:


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -370,114 +393,76 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+                    // If we get None here, it means the child node is deleted(inconsistent)
+                    // and we should retry
+                    let value = match option_value {
+                        Some(value) => value,
+                        None => {
+                            debug!(
+                                "lookup() found the child ino={} is deleted(inconsistent)",
+                                ino
+                            );
+                            continue;
```
ActivePeter, line 424: Seems these two continue should directly return error. The first points out the inconsistency between the metadata information of the parent node and the key-value (kv) of the child node. This type of inconsistency may require centralized statistical analysis and discussion. The second "continue" is usually an indication of an error in etcd. It should return the error directly.

xiaguan: "I think you're right about the first point. However, for the second one, I believe it might be better to try again. Maybe this error can be resolved by retrying, especially for etcd.

Issue #379, opened 2023-05-29
Add：Distribute id allocator


ActivePeter:
# Background:

The previous inode number allocation module was specifically designed to allocate inodes using a pre-allocation method to reduce the frequency of remote calls. When implementing file operation locking, it was discovered that the allocation of file descriptors (fd) also falls under the global ID allocation. Therefore, this module was separated and made into a reusable module. 

The original design of Inode number allocation was as follows: https://github.com/datenlord/datenlord/issues/349

The basic idea was to use a global key-value store to obtain an incrementing value. However, using the key-value store for every increment would incur high remote call overhead and competition. To mitigate this, a pre-allocation method can be used by incrementing a range of values. After the initial allocation, subsequent allocations can use values from within the range until it is exhausted. Only then, remote operations are performed to obtain a new range.

However, since this module was not designed with reusability in mind, the naming and logic were tightly coupled with inodes. Therefore, this design needs to separate it to facilitate reuse.

# Overall Approach:

- Add the id_alloc module to memfs::dist, implementing IdAllocator.
- Replace the kv part from the previous etcd_delegate to kv_engine.
- Different IDs require different keys for value storage and locking protection. Introduce a design similar to KeyType in KVEngine, introducing IdType and implementing a to_str function to obtain the key's prefix. Also, add the necessary Key types for IdAllocator in KeyType.
- Since this module requires lock operations, as well as regular get, set, and unlock operations, implement the get, set, lock, and unlock methods in KVEngine.
- For different specific ID allocators, the main initialization parameters required are IdType and the beginning number.
Since inum alloc and fd alloc involve not only incrementing values but also additional operations, specific implementations are required.

# Specific Implementations:

kv_engine.rs
- Add lock, unlock, get, set, and delete interfaces to support non-transactional usage.
- Add LockKeyType to standardize the key type used for locks in the entire system.
- Add the key and value types for the corresponding ID allocator in LockKeyType, KeyType, and ValueType.
i- d_alloc.rs
```rust
pub enum ValueType {
    ...
    /// Next id allocate range begin
    NextIdAllocateRangeBegin(u64),
}
pub enum KeyType<'a> {
    ...
    /// IdAllocator value key
    IdAllocatorValue {
        /// the prefix of the key for different id type
        prefix: &'a str,
    },
}
pub enum LockKeyType<'a> {
    ...
    /// IdAllocator lock key
    IdAllocatorLock {
        /// the prefix of the key for different id type
        prefix: &'a str,
[...]

PR #380, opened 2023-05-29
Add: Distrubuted id allocator

ActivePeter:
The design issue: https://github.com/datenlord/datenlord/issues/379


xiaguan:
MetaTxn() does not require a cancel() interface as it is designed for the read-modify-writeback model. If you receive an unwanted value during a get() operation, you can simply return an Error or use KVEngine::get().

ActivePeter:
> `MetaTxn` does not require a `cancel()` interface as it is designed for the read-modify-writeback model. If you receive an unwanted value during a `get()` operation, you can simply return an Error or use `KVEngine::get()`.

The purpose here is to make retry_txn macro more convenient. In some cases we want to break the retry loop but we don't want to commit the txn, then we can call this function. Just check the code used this function.

xiaguan:
ID allocation shouldn't require locks.Remove the content related to lock impl.

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -20,66 +23,127 @@ pub enum ValueType {
...
+    /// # Panics
     /// Panics if `ValueType` is not `ValueType::Node`.
     pub async fn into_s3_node<S: S3BackEnd + Send + Sync + 'static, K: KVEngine + 'static>(
         self,
         meta: &S3MetaData<S, K>,
     ) -> S3Node<S> {
         match self {
             ValueType::Node(node) => S3Node::from_serial_node(node, meta).await,
-            ValueType::DirEntry(_) | ValueType::INum(_) | ValueType::Attr(_) => {
+            _ => {
```
rogercloud, line 41: We'd better list all the possible values. If we add new enum option later, we may handle it incorrectly as there's no error reporting for the wildcard_enum_match.

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -20,66 +23,127 @@ pub enum ValueType {
...
-    #[allow(dead_code)]
+    /// Turn the `ValueType` into `NextIdAllocateRangeBegin`.
+    /// # Panics
+    /// Panics if `ValueType` is not `ValueType::NextIdAllocateRangeBegin`.
+    #[allow(clippy::wildcard_enum_match_arm)]
     #[must_use]
+    pub fn into_next_id_allocate_range_begin(self) -> u64 {
+        match self {
+            ValueType::NextIdAllocateRangeBegin(begin) => begin,
+            _ => panic!("expect ValueType::NextIdAllocateRangeBegin but get {self:?}"),
```
rogercloud, line 55: The same as above.

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -20,66 +23,127 @@ pub enum ValueType {
...
-#[derive(Debug)]
 /// The `KeyType` is used to locate the value in the distributed K/V storage.
 /// Every key is prefixed with a string to indicate the type of the value.
 /// If you want to add a new type of value, you need to add a new variant to the enum.
 /// And you need to add a new match arm to the `get_key` function , make sure the key is unique.
 #[allow(dead_code)]
-pub enum KeyType {
-    /// INum -> SerialNode
+#[derive(Debug, Eq, PartialEq)]
+pub enum KeyType<'a> {
```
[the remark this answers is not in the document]
ActivePeter: It's for str ref.

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -20,66 +23,127 @@ pub enum ValueType {
...
+impl KeyType<'_> {
     /// Get the key in bytes.
+    #[must_use]
     pub fn get_key(&self) -> Vec<u8> {
         match *self {
-            KeyType::INum2Node(ref i) => format!("I{i}").into_bytes(),
-            KeyType::INum2DirEntry(ref i) => format!("D{i}").into_bytes(),
-            KeyType::Path2INum(ref p) => format!("P{p}").into_bytes(),
-            KeyType::INum2Attr(ref i) => format!("A{i}").into_bytes(),
+            KeyType::INum2Node(ref i) => format!("NodeI{i}").into_bytes(),
```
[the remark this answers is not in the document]
ActivePeter: This is just an encoding problem, if we use one litter prefix, there‘s only 26*2 key prefixes can be used

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -193,6 +258,9 @@ impl MetaTxn for EtcdTxn {
     }
 
     async fn commit(&mut self) -> DatenLordResult<bool> {
+        if self.version_map.is_empty() && self.buffer.is_empty() {
```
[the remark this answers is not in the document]
ActivePeter: Just for improving code completeness.

ActivePeter:
> 

Why?

ActivePeter:
> ID allocation shouldn't require locks.Remove the content related to lock impl.

Why?

src/async_fuse/memfs/kv_engine.rs
ActivePeter: These functions are for turning a common type (ValueType) into one certained type (S3_Node), if later we implement more types, would that be too long or weird? And actually we always are certain a key type is binded with a value type.

src/async_fuse/memfs/kv_engine.rs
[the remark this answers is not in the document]
ActivePeter: We don't need to commit a transaction if there's no operation.

src/async_fuse/memfs/kv_engine.rs
ActivePeter: Added comment for this.

src/async_fuse/memfs/kv_engine.rs
ActivePeter: Added comment for this in code.

xiaguan (approved):

Issue #373, opened 2023-05-10
Add distributed rwlock implemented by kv storage


ActivePeter:
# Background

When implementing file read-write locking mechanisms, the remote read-write lock functionality is required. However, etcd currently does not have read-write lock functionality, so it needs to be implemented using transactions.

# Reference

- [etcd-lock/rwlock.go at master · diamanticom/etcd-lock](https://github.com/diamanticom/etcd-lock/blob/master/rwlock.go)

# Basic Introduction

Here we need to implement a reentrant distributed read-write lock with a timeout feature, which automatically releases the lock if the timeout is exceeded.

Reentrant: If the lock has already been acquired, as long as it is the same node and the same type of lock, it can be successfully acquired again.

# Location of mod

Since this functionality is reusable, the module will be placed under `common::dist_rwlock`.

# pub api

``` rust
pub async fn rw_lock(
    etcd: &EtcdDelegate,
    name: &str,
    locktype: DistRwLockType,
    timeout: Duration,
    tag_of_local_node: &str, // mark node tag
) -> DatenLordResult<()> 

pub async fn rw_unlock(
    etcd: &EtcdDelegate,
    name: &str,
    tag_of_local_node: &str, // mark node tag
) -> DatenLordResult<()> 
```

# Lock Process
```
Loop:
   Attempt to write lock the key-value (KV).
   If successful:
      Return.
   Else:
      Lock already exists.
      If requesting read (r) and the current lock is read (r):
         Add current node data to the lock information, renew the lock lease, and return.
      If requesting write (w) and the current lock is write (w):
         If it's the same node, renew the lock lease and return.
         If it's a different node, wait for the lock to be released and go back to the loop.
      If requesting write (w) and the current lock is read (r) or requesting read (r) and the current lock is write (w):
         Wait for the lock to be released.
         Go back to the loop.
```
# Unlock Process
```
Retrieve lock data:
   If the lock exists, remove the current node from the current lock information:
      If no other nodes hold the lock:
         Remove the key.
      If other nodes hold the lock:
         Update the lock ownership information (requires a transaction, retry on failure).
```

PR #381, opened 2023-05-29
Refactor: return version in etcd_delegate

ActivePeter:
When implement distribute rwlock, a version for etcd txn is needed to be return.

### Relate:
- dist rwlock design https://github.com/datenlord/datenlord/issues/373


src/common/etcd_delegate.rs
```diff
@@ -101,7 +106,7 @@ impl EtcdDelegate {
     async fn get_one_kv_async<T: DeserializeOwned, K: Into<Vec<u8>> + Debug + Clone + Send>(
         &self,
         key: K,
-    ) -> DatenLordResult<Option<T>> {
+    ) -> DatenLordResult<Option<(T, KVVersion)>> {
```
xiaguan, line 109: Maybe we can add a separate function called "get_with_version" because most of the time when we call this interface, we don't need the version. From your pull request, it seems to be causing issues with unused variables and making the code less aesthetically pleasing with added underscores.

src/common/etcd_delegate.rs
```diff
@@ -111,12 +116,96 @@ impl EtcdDelegate {
...
+                util::decode_from_bytes::<T>(kv.value())?,
+                kv.version(),
+            ))),
             None => Ok(None),
         }
     }
 
+    /// offer an version check
+    async fn write_to_etcd_with_version<
+        T: DeserializeOwned + Serialize + Clone + Debug + Send + Sync,
```
xiaguan, line 131: Just a suggestion: Perhaps we should replace these template constraints with &[u8] instead. This would result in faster compilation speed since templates wouldn't be required.

src/common/etcd_delegate.rs
```diff
@@ -111,12 +116,96 @@ impl EtcdDelegate {
...
+                            "failed to get LeaseGrantResponse from etcd, the timeout={}",
+                            dur.as_secs()
+                        )
+                    })?
+                    .id(),
+            );
+        };
+        let txn_req = etcd_client::EtcdTxnRequest::new()
+            // etcd：chack key exist in txn
+            //  https://github.com/etcd-io/etcd/issues/7115
```
xiaguan, line 160: It would be better to include this suggestion in the official documentation (https://etcd.io/docs/v3.4/learning/api/).

src/common/etcd_delegate.rs
```diff
@@ -111,12 +116,96 @@ impl EtcdDelegate {
...
+        prev_version: KVVersion,
+        timeout: Option<Duration>,
+    ) -> DatenLordResult<Option<(T, KVVersion)>> {
+        let bin_value = bincode::serialize(value)
+            .with_context(|| format!("failed to encode {value:?} to binary"))?;
+        let mut put_request = etcd_client::EtcdPutRequest::new(key.clone(), bin_value);
+        if let Some(dur) = timeout {
+            put_request.set_lease(
+                self.etcd_rs_client
+                    .lease()
```
xiaguan, line 146: Why does the 'put' operation here require a lease() to ensure?

src/common/etcd_delegate.rs
```diff
@@ -111,12 +116,96 @@ impl EtcdDelegate {
...
+            .etcd_rs_client
+            .kv()
+            .txn(txn_req)
+            .await
+            .with_context(|| {
+                format!("failed to get PutResponse from etcd for key={key:?}, value={value:?}",)
+            })?;
+
+        if txn_res.is_success() {
+            //key does not exist, insert kv
```
xiaguan, line 184: Is the comment written incorrectly here?

xiaguan (changes requested):

src/common/etcd_delegate.rs
ActivePeter: It will be replace with kvengine, we don't need to waste too much effort on this.

src/common/etcd_delegate.rs
ActivePeter: I think that's not necessary.

src/common/etcd_delegate.rs
ActivePeter: Lease is for timeout.

src/common/etcd_delegate.rs
ActivePeter: What suggestion?

src/common/etcd_delegate.rs
ActivePeter: And actully most part need to use txn or lock finally.

xiaguan: But this modification includes some unnecessary changes in this pull request, such as csi/meta, and causes some aesthetic issues. If you need get_with_version, you just need to create a new function

src/common/etcd_delegate.rs
xiaguan: So why does the `put` operation here require a timeout?

ActivePeter: That's common, sometimes a value is just for implement lock, avoid conflict operation by txn or temporary use, we don't want to keep it in kv all the time.

scripts/coverage.sh
```diff
@@ -12,6 +12,14 @@ else
   OUTPUT_DATENLORD=coverage/datenlord_cov.lcovrc
 fi
 
+# install and start etcd
```
rogercloud, line 15: It's weird to start etcd service in the coverage script. I get the point that we need to get test result with the service on, but it's also not a practice to make `cargo test` depend on a third-party service.

Usually we use mock service in the test to make it clean.

src/async_fuse/memfs/dist/etcd.rs
```diff
@@ -1,6 +1,7 @@
 use crate::async_fuse::memfs::INum;
 use crate::common::error::Context;
 use crate::common::etcd_delegate::EtcdDelegate;
+use datenlord::common::etcd_delegate::KVVersion;
```
rogercloud, line 4: I prefer `KvVersion` instead of `KVVersion`.

src/async_fuse/memfs/dist/etcd.rs
```diff
@@ -60,7 +61,7 @@ pub async fn get_node_ip_and_port(
             format!("Fail to get Node Ip address and port information, node_id:{node_id}")
         })?;
 
-    if let Some(ip_and_port) = ip_and_port {
+    if let Some((ip_and_port, _v)) = ip_and_port {
```
rogercloud, line 64: If we usually drop the version, why don't we provide a function to drop it and return the value alone?

src/common/etcd_delegate.rs
rogercloud: Currently it seems not necessary, but we use generic to make it extendable. So I prefer to keep it that way.

scripts/coverage.sh
ActivePeter: Ok. I'll change relate test to mock.

src/async_fuse/memfs/dist/etcd.rs
ActivePeter: OK.

src/async_fuse/memfs/dist/etcd.rs
ActivePeter: Then we need a fn get_at_most_one_value and fn get_at_most_one_value_without_version. I think the difference is very small, and the copy of version cost very little.

ActivePeter: Changed to related functions to fn xxxx and fn xxxx_with_version, and the default xxxx returns withouts version.

src/common/etcd_delegate.rs
ActivePeter: Changed to related functions to fn xxxx and fn xxxx_return_with_version, and the default xxxx returns withouts version.

src/async_fuse/memfs/dist/etcd.rs
```diff
@@ -156,12 +157,12 @@ pub async fn get_volume_nodes(
     volume_info: &str,
 ) -> anyhow::Result<HashSet<String>> {
     let volume_info_key = format!("{ETCD_VOLUME_INFO_PREFIX}{volume_info}");
-    let volume_node_list: Option<Vec<u8>> = etcd_client
+    let volume_node_list: Option<(Vec<u8>, KvVersion)> = etcd_client
```
xiaguan, line 160: Did you forget to modify this part?

ActivePeter: Ops

xiaguan (approved):

PR #389, opened 2023-06-25
Upgrade Project Code Coverage Testing

xiaguan:
This pull request introduces an upgrade to our code coverage testing in the CI pipeline.


codecov.info
rogercloud, line 1: What's the purpose of this file? I don't think we should check it in.

PR #390, opened 2023-06-29
Add lock manager

ActivePeter:
It can auto renew the lock lease before the lock expires.
So the lock is held until out client close or call the unlock operation.

Offer some lock apis, will offer rwlock in the future.
- trylock
- lock
- unlock

Do the keep alive lease job when waiting for lock.


src/async_fuse/memfs/dist/lock_manager.rs
```diff
@@ -0,0 +1,384 @@
...
+        let key_vec: Vec<u8> = key.get_key();
+
+        // already locked, only need the system to renew the lease
+        if self.locked.read().timeout_renews.contains_key(&key_vec) {
+            return Ok(true);
+        }
+        let lease_info = self.kv_engine.lease_grant(DEFAULT_TIMEOUT).await?;
+        // lock
+        let lockinfo = if try_lock {
+            match tokio::time::timeout(
```
xiaguan, line 170: When locking a key in etcd, a k/v pair is generated which can be retrieved using 'get'. If the key already exists, 'try_lock' will return immediately. This can reduce the delay caused by 'try_lock'.

src/async_fuse/memfs/dist/lock_manager.rs
```diff
@@ -0,0 +1,384 @@
...
+            test_lockmanager_once().await;
+        }
+    }
+
+    // #[tokio::test(flavor = "multi_thread")]
+    #[allow(clippy::unwrap_used, clippy::too_many_lines)]
+    async fn test_lockmanager_once() {
+        common::util::setup_test_log_debug();
+
+        let mut server = mock_etcd::MockEtcdServer::default();
```
xiaguan, line 287: We can move these preparation tasks to another function, which may be reused in the future. This also keeps the testing logic clean.

xiaguan (changes requested):

xiaguan:
Because this is a very important new feature for the future work, we must ensure its correctness as well as performance. The current test code has a high degree of code repetition, so we can extract some functions such as `lock(expect_succ: bool)`. At the same time, we need a more diverse set of random tests. As for performance testing, we can print logs in these tests to display the latency.

xiaguan:
> Because this is a very important new feature for the future work, we must ensure its correctness as well as performance. The current test code has a high degree of code repetition, so we can extract some functions such as `lock(expect_succ: bool)`. At the same time, we need a more diverse set of random tests. As for performance testing, we can print logs in these tests to display the latency.

https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/mutex_test.cc， this test about locks is well-writen. We can take reference from it.

src/async_fuse/memfs/kv_engine.rs
```diff
@@ -401,7 +491,7 @@ macro_rules! retry_txn {
 }
 
 #[async_trait]
-impl KVEngine for EtcdKVEngine {
+impl KVEngine<EtcdLeaseKeeper> for EtcdKVEngine {
```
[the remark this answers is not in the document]
ActivePeter: Done.

src/async_fuse/memfs/dist/lock_manager.rs
ActivePeter: image

Later we can test the self implemented rwlock's performance, if it's better we can replace the native lock with it.

src/async_fuse/memfs/dist/lock_manager.rs
ActivePeter: Agree with that.

PR #391, opened 2023-07-12
Fix k8s test

ActivePeter:
Github action environment has too many preload tools that we don’t need which takes too many space, after removed them, the cron test passed.
- Add some rm operation to cron.yml to get enough space for test running.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -35,7 +35,7 @@ env:
   K8S_CONFIG: k8s.e2e.config
   K8S_VERSION: v1.21.1
   KIBANA_LABEL: kibana
-  KIND_NODE_VERSION: kindest/node:v1.21.1
+  KIND_NODE_VERSION: kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6
   KIND_VERSION: v0.11.1
   # MINIKUBE_VERSION: v1.13.0
   NODE_APP_LABEL: csi-nodeplugin-datenlord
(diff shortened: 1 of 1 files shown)


scripts/local_test/README.md
```diff
@@ -0,0 +1,70 @@
+# Operating requirement
+
+- Run the scripts with bash, because we need to load env variables from cron.yml
+
+- Run the third script in common user, the others in root
```
xiaguan, line 5: Providing a specific file name may be better?

scripts/local_test/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Run 4.1 after `cargo build --release`
+
+- If there's problem after 4.2, just check your proxy setting and switch proxy node.
+
+- Make sure the virtual machine has enough resouces, 8g mem, 4 cores, 200g disk.
+
+# Practical steps
+
+1. Create a vm for test. Install some basic software: ssh, vim, cmake, protobuf, rust, curl, take snapshot for the vm as Snapshot1
```
xiaguan, line 15: Can we write a script for setting up the development environment based on the information in `ci.yml`?

scripts/local_test/README.md
```diff
@@ -0,0 +1,70 @@
...
+2. Run 1-3 scripts
+  
+   When running 2, you should make sure the network is ok, if failed the config the proxy like following guidance. If after all these done and you still failing, just roll back tp SnapShot1
```
xiaguan, line 18: Providing a specific file name may be better?

scripts/local_test/README.md
```diff
@@ -0,0 +1,70 @@
...
+- Make sure the virtual machine has enough resouces, 8g mem, 4 cores, 200g disk.
+
+# Practical steps
+
+1. Create a vm for test. Install some basic software: ssh, vim, cmake, protobuf, rust, curl, take snapshot for the vm as Snapshot1
+2. Run 1-3 scripts
+  
+   When running 2, you should make sure the network is ok, if failed the config the proxy like following guidance. If after all these done and you still failing, just roll back tp SnapShot1 
+
+   When running 3, you should make sure you run it in common user and give the user docker permission as following.
```
xiaguan, line 20: Providing a specific file name may be better?

scripts/local_test/README.md
```diff
@@ -0,0 +1,70 @@
...
+  ``` bash
+  sudo gpasswd -a $USER docker  #将登陆用户加入到docker用户组中
+  newgrp docker #更新用户组
+  ```
+
+# Proxy may need
+
+- curl
+  ```
+  export http_proxy="http://192.168.232.1:23333"
```
xiaguan, line 41: Use `your_proxy_server_url` instead.

xiaguan:
I think we have too many scripts. We should try to reduce the number of scripts by using conditional statements or merging scripts that do not conflict with each other.

scripts/local_test/README.md
ActivePeter: That's not necessary because different systems may have different install way. And a snapshot will be token after them installed.

rogercloud:
I recommend to move the local scripts to another PR. And I agree with @xiaguan that we should reduce the number of scripts files, otherwise it's hard to maintain.

ActivePeter:
> 

Ok.

rogercloud (approved):

PR #394, opened 2023-07-13
Add local test scripts.

ActivePeter:
There's proxy or environment problems happens when doing local test for k8s environment.
This pr offers some scripts for local k8s test.


Reto911:
Already done in #406. But we need to add some documents for running tests locally.

PR #406, opened 2023-09-06
cron-ci: decompose cron tests into independent scripts

GTwhy:


xiaguan:
I think the script itself is well organized. But it still feels a bit messy🤔. https://github.com/datafuselabs/databend/blob/main/scripts/, I think this is a very good reference.Use folders to categorize scripts of the same type.

.travis.yml
xiaguan, line 1: Since we are not using Travis CI, we can remove the related files.

docs/datenlord_monitoring.md
```diff
@@ -49,7 +49,7 @@ Use the following default username and password to log in. Once you log in with
 User: admin
 Pass: admin
 ```
-Datenlord has already built a dashboard templates. To use Datenlord dashboard, just import [datanlord_dashboard.json](../scripts/datenlord-e2e-test.yaml) file in Grafana.
+Datenlord has already built a dashboard templates. To use Datenlord dashboard, just import [datanlord_dashboard.json](../scripts/setup/alertmanager-alerts.yaml) file in Grafana.
```
xiaguan, line 52: Is this a mistake?

scripts/ci/datenlord-monitor-test.sh
```diff
@@ -31,7 +30,8 @@ kind load docker-image prom/alertmanager:$ALERTMANAGER_VERSION
 kind load docker-image docker.elastic.co/elasticsearch/elasticsearch-oss:$ELASTIICSEARCH_VERSION
 kind load docker-image docker.elastic.co/kibana/kibana-oss:$KIBANA_VERSION
 kind load docker-image prom/node-exporter
-sh scripts/datenlord-monitor-deploy.sh deploy
+sudo chmod +x scripts/setup/datenlord-monitor-deploy.sh
```
xiaguan, line 33: You can set all scripts to be executable locally, so there's no need to use chmod

scripts/setup/setup-ssh-for-cluster-nodes.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+source scripts/setup/config.sh
+
+rm -f /tmp/sources.list
+cat > /tmp/sources.list <<'END'
+# ubuntu20
+deb http://mirrors.aliyun.com/ubuntu/ focal main restricted
```
xiaguan, line 8: I believe the CI on GitHub runs on AWS, so there's no need to change the `sources.list`.

xiaguan (changes requested):

docs/datenlord_monitoring.md
GTwhy: Yes, the original `e2e-test` is a mistake, and `alerts` is as well.
I think the `Datenlord-Dashboard.json` is the right one.

scripts/setup/setup-ssh-for-cluster-nodes.sh
GTwhy: Yes, we will run CI on GIthub or AWS, but these scripts can all be reused in local testing, and aliyun may be more friendly for Chinese users.

xiaguan (dismissed):

xiaguan (dismissed):

scripts/ci/datenlord-monitor-test.sh
```diff
@@ -13,8 +11,9 @@ ELASTIICSEARCH_VERSION=${ELASTIICSEARCH_VERSION:-"7.4.2"}
 KIBANA_VERSION=${KIBANA_VERSION:-"7.2.0"}
 
 # Datenlord Monitoring test
-sudo chmod +x scripts/datenlord-monitor-deploy.sh
-cat scripts/alertmanager_test_alert.yaml >> scripts/alertmanager_alerts.yaml
+if ! grep -q "alertmanager_demo" scripts/setup/alertmanager-alerts.yaml; then
+  cat scripts/ci/alertmanager-test-alert.yaml >> scripts/setup/alertmanager-alerts.yaml
+fi
```
rogercloud, line 16: Shall we merge these two files into a new one? If so, the users don't have to reset the file after the script.

rogercloud (commented):
I see that you replace sh with bash in all scripts. Though bash is better than sh, do you consider the system compatibility?

GTwhy:
> I see that you replace sh with bash in all scripts. Though bash is better than sh, do you consider the system compatibility?

The compatibility of sh is better, we can use sh as default,  unless we need the feature of bash.

xiaguan (approved):

PR #413, opened 2023-09-26
fix: Fix sticky bit related bugs

xiaguan:
Fixed a bug related to the sticky bit. Added tests` test_create_dir` and `test_create_symlink` because this bug was caused by the improper creation of symlinks. No tests related to the sticky bit were included, as it is difficult to simulate user switching in Rust tests. Relevant tests will be covered in `fstest`.

### DIFF
--- a/src/async_fuse/memfs/s3_metadata.rs
+++ b/src/async_fuse/memfs/s3_metadata.rs
@@ -550,11 +550,11 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
             Some(ref path) => Some(path.as_ref()),
             None => None,
         };
-        let uid = param.uid;
-        let gid = param.gid;
+        let user_id = param.uid;
+        let group_id = param.gid;
         // pre-check : check whether the child name is valid
         let mut parent_node = self
-            .create_node_pre_check(parent, node_name, uid, gid)
+            .create_node_pre_check(parent, node_name, user_id, group_id)
             .await
             .context("create_node_helper() failed to pre check")?;
         // allocate a new i-node number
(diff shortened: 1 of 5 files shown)


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -1168,6 +1170,7 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
     }
 
     /// Rename helper function to pre-check
+    #[allow(clippy::too_many_lines)]
```
rogercloud, line 1173: Do we have an issue to track this function refactor?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -1521,6 +1535,7 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
     }
 
     /// Helper function to remove node locally
+    #[allow(clippy::too_many_lines)]
```
rogercloud, line 1538: Same as above.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -550,11 +550,11 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
             Some(ref path) => Some(path.as_ref()),
             None => None,
         };
-        let uid = param.uid;
-        let gid = param.gid;
+        let user_id = param.uid;
```
rogercloud, line 553: I think `uid` and `gid` is accurate in this case. The naming is a kind of convention.

Reto911:
Done by another PR.

PR #420, opened 2023-10-23
Make s3 key Inum

rogercloud:
This commit does two things:

1. change s3 persistent storage's key from full path to Inum.

2. Remove S3 async task part.

The reason to remove S3 async task is that we don't need to store metadata to S3 storage anymore.


xiaguan (approved):
LGTM

Issue #376, opened 2023-05-17
Excessive Usage of panic! in memfs Metadata Files


xiaguan:
I've been observing the use of the `panic!` macro in our memfs metadata files and I believe it is being overused in some scenarios.

While `panic!` is appropriate for situations where the program absolutely cannot continue to operate, there are instances in our code where we could gracefully handle the error and possibly recover instead of panicking.The usage of `panic!` could lead to abrupt termination of the program, making it hard to debug and causing potential downtime in production environments. It's better practice to return Result types which give us an opportunity to handle errors in a controlled manner.
I propose that we revisit these files and refactor the code to minimize the usage of panic!, and where possible, replace it with error handling mechanisms.

Reto911:
All problems mentioned in this issue are resolved, therefore the issue can be closed now.

PR #421, opened 2023-10-25
Fix: Remove Excessive Usage of `panic!` in memfs Metadata Files

Reto911:
This PR aims to fix a problem of excessive usage of `panic!` in `async_fuse/memfs/s3_metadata.rs`, raised by #376 .

### DIFF
--- a/src/async_fuse/memfs/s3_metadata.rs
+++ b/src/async_fuse/memfs/s3_metadata.rs
@@ -13,7 +13,7 @@ use itertools::Itertools;
 use nix::errno::Errno;
 use nix::fcntl::OFlag;
 use nix::sys::stat::SFlag;
-use parking_lot::RwLock as SyncRwLock;
+use parking_lot::RwLock as SyncRwLock; // conflict with tokio RwLock
 use tokio::sync::Mutex;
 use tracing::{debug, error, instrument, warn};
 
@@ -24,6 +24,7 @@ use super::dist::server::CacheServer;
 use super::fs_util::{self, FileAttr};
 use super::id_alloc_used::INumAllocator;
 use super::kv_engine::{KVEngine, KVEngineType, KeyType, ValueType};
+use super::metadata::error;
 use super::metadata::{MetaData, ReqContext};
 use super::node::Node;
 use super::s3_node::S3Node;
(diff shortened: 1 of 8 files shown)


Reto911:
The usage of `panic!` calls has been reduced from 52 to 1, and the remain one represents an unimplemented function.

https://github.com/Reto911/datenlord/blob/e552dae60febf4a9f9e150363dfdffe58d8dff6e/src/async_fuse/memfs/s3_metadata.rs#L1330

Reto911:
Besides, as some caller that returns `nix::Result` but not `DatenLordResult`, the returned `DatenLordResult` cannot be passed up. Therefore, those caller that returns `nix::Result` still panics when get an `Error`. There are 4 such `panic!` calls.

As `nix::Result` can be converted into `DatenLordResult`, this might be fixed in the future,

xiaguan:
The content of the PR is much appreciated. However, please ensure that the commit titles adhere to the Conventional Commits specification, which you can find here: [Conventional Commits 1.0.0](https://www.conventionalcommits.org/en/v1.0.0/). The title of the PR should follow the same standard. We could consider using a CI tool to automatically enforce this.

xiaguan (dismissed):

Reto911:
> The content of the PR is much appreciated. However, please ensure that the commit titles adhere to the Conventional Commits specification, which you can find here: [Conventional Commits 1.0.0](https://www.conventionalcommits.org/en/v1.0.0/). The title of the PR should follow the same standard. We could consider using a CI tool to automatically enforce this.

Thank you for your reminding! I will pay attention on it in the future.

xiaguan:
You can edit the commit messages and then force-push. It's alright.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -68,24 +72,29 @@ pub struct S3MetaData<S: S3BackEnd + Send + Sync + 'static> {
 }
 
 /// Parse S3 info
-fn parse_s3_info(info: &str) -> (&str, &str, &str, &str) {
+fn parse_s3_info(info: &str) -> DatenLordResult<(&str, &str, &str, &str)> {
     info.split(S3_INFO_DELIMITER)
         .next_tuple()
-        .unwrap_or_else(|| panic!("parse s3 information failed. s3_info: {info}"))
+        .ok_or_else(|| anyhow::anyhow!("parse s3 information failed. s3_info: {info}").into())
```
rogercloud, line 78: It's a very general input parsing error, we can add an error type for it, or maybe we've already have one.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -68,24 +72,29 @@ pub struct S3MetaData<S: S3BackEnd + Send + Sync + 'static> {
...
+        ino: u64,
+        fh: u64,
+        _flags: u32,
+        _lock_owner: u64,
+        flush: bool,
+    ) -> DatenLordResult<()> {
+        let mut inode = self
+            .get_node_from_kv_engine(ino)
+            .await?
+            .ok_or_else(|| error::build_inconsistent_fs(ino, "release"))?;
```
rogercloud, line 97: Maybe try https://stackoverflow.com/questions/38088067/equivalent-of-func-or-function-in-rust this way to get the function name.

src/async_fuse/memfs/s3_node.rs
```diff
@@ -155,7 +155,11 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3Node<S> {
                 for (name, serial_dir_entry) in dir {
                     let child_ino = serial_dir_entry.get_child_ino();
                     // Fetch the child node from kv
-                    let child_node = meta.get_node_from_kv_engine(child_ino).await;
+                    // TODO: Handle this Error
```
rogercloud, line 158: Please explain the reason to handle this later.

Reto911: I am sorry that this is my negligence, and it's fixed now.

src/async_fuse/memfs/s3_metadata.rs
Reto911: Thank you for your suggestion. It's applied in my code now.

src/async_fuse/memfs/s3_metadata.rs
Reto911: We do have one, called `S3Error`, but it's not designed well. We can refactor it in the future.

rogercloud (approved):

PR #423, opened 2023-11-07
feat: implement exchange rename and remove some dead code

Reto911:
As "exchange rename" has not yet been implemented, this PR implements it.

There are 3 main proposes in this PR:
1. Remove some dead code about `full_path` from `S3Node`, `S3Metadata` and etcd. 
2. Implement "exchange rename", which can exchange 2 files (may be in different types) atomically, and write an integration test case for it.
3. Mark codes related to FUSE request `Rename2` with feature `abi-7-23`, as `Rename2` is not supported in protocol version lower than 7.23.

Besides, this PR introduces a CI workflow for protocol version 7.23.

### DIFF
--- a/scripts/setup/start_local_node.sh
+++ b/scripts/setup/start_local_node.sh
@@ -1,5 +1,13 @@
 #!/bin/sh
 
+# scripts/setup/start_local_node.sh [build_flags]
+#
+# The parameter `build_flags` will be passed to `cargo build` in this scripts
+# when building DatenLord. You should put all options in quotes.
+#
+# For example, run `scripts/setup/start_local_node.sh`, this script will just simply run `cargo build` without any option.
+# But run `scripts/setup/start_local_node.sh "-F abi-7-23"`, this script will run `cargo build -F abi-7-23` to build DatenLord.
+
 export CONTROLLER_SOCKET_FILE=/tmp/controller.sock
 export BIND_MOUNTER=../target/debug/bind_mounter
 export NODE_SOCKET_FILE=/tmp/node.sock
@@ -8,6 +16,9 @@ export RUST_LOG=debug
 export RUST_BACKTRACE=1
 export ETCD_END_POINT=127.0.0.1:2379
 export BIND_MOUNTER=`realpath $BIND_MOUNTER`
+
+# build flags with `cargo build`
+BUILD_FLAGS=$1
 
 . scripts/setup/config.sh
 . scripts/setup/setup_etcd.sh
@@ -32,7 +43,7 @@ fi
 
 echo "==> Start to deploy datenlord locally"
 echo "==> Building datenlord"
-cargo build
+cargo build $BUILD_FLAGS
 if [ $? -ne 0 ]; then
   echo "Failed to build datenlord."
   exit 1
--- a/src/async_fuse/memfs/kv_engine/mod.rs
+++ b/src/async_fuse/memfs/kv_engine/mod.rs
@@ -25,6 +25,9 @@ pub mod etcd_impl;
 pub mod kv_utils;
 
 /// The `ValueType` is used to provide support for metadata.
+///
+/// The variants `DirEntry`, `INum` and `Attr` are not used currently,
+/// but preserved for the future.
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub enum ValueType {
     /// SerialNode
@@ -120,8 +123,6 @@ pub enum KeyType {
     INum2Node(INum),
     /// INum -> DirEntry
     INum2DirEntry(INum),
-    /// Path -> Inum
-    Path2INum(String),
     /// INum -> SerialFileAttr
     INum2Attr(INum),
     /// IdAllocator value key
@@ -135,6 +136,9 @@ pub enum KeyType {
     /// Node list
     /// The corresponding value type is ValueType::RawData
     FileNodeList(INum),
+    /// Just a string key for testing the KVEngine.
+    #[cfg(test)]
+    String(String),
 }
 
 // ::<KeyType>::get() -> ValueType
@@ -155,7 +159,8 @@ impl Display for KeyType {
         match *self {
             KeyType::INum2Node(ref i) => write!(f, "INum2Node{{i: {i}}}"),
             KeyType::INum2DirEntry(ref i) => write!(f, "INum2DirEntry{{i: {i}}}"),
-            KeyType::Path2INum(ref p) => write!(f, "Path2INum{{p: {p}}}"),
+            #[cfg(test)]
+            KeyType::String(ref s) => write!(f, "String{{s: {s}}}"),
             KeyType::INum2Attr(ref i) => write!(f, "INum2Attr{{i: {i}}}"),
             KeyType::IdAllocatorValue(ref id_type) => {
                 write!(f, "IdAllocatorValue{{unique_id: {id_type:?}}}")
@@ -205,7 +210,8 @@ impl KeyType {
         match *self {
             KeyType::INum2Node(ref i) => serialize_key(0, i),
             KeyType::INum2DirEntry(ref i) => serialize_key(1, i),
-            KeyType::Path2INum(ref p) => serialize_key(2, p),
+            #[cfg(test)]
+            KeyType::String(ref s) => serialize_key(2, s),
             KeyType::INum2Attr(ref i) => serialize_key(3, i),
             KeyType::IdAllocatorValue(ref id_type) => serialize_key(4, &id_type.to_unique_id()),
             KeyType::NodeIpPort(ref s) => serialize_key(6, s),
--- a/src/async_fuse/memfs/s3_metadata.rs
+++ b/src/async_fuse/memfs/s3_metadata.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, VecDeque};
+use std::collections::BTreeMap;
 use std::os::unix::ffi::OsStringExt;
 use std::os::unix::io::RawFd;
 use std::path::Path;
@@ -15,7 +15,7 @@ use nix::fcntl::OFlag;
 use nix::sys::stat::SFlag;
 use parking_lot::RwLock as SyncRwLock; // conflict with tokio RwLock
 use tokio::sync::Mutex;
-use tracing::{debug, error, instrument, warn};
+use tracing::{debug, instrument, warn};
 
 use super::cache::{GlobalCache, IoMemBlock};
 use super::dir::DirEntry;
@@ -453,12 +453,8 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
         let root_inode = S3Node::open_root_node(FUSE_ROOT_ID, "/", s3_backend, Arc::clone(&meta))
             .await
             .add_context("failed to open FUSE root node")?;
-        let full_path = root_inode.full_path().to_owned();
-        // insert two K/V pairs into KV engine
-        // 1. FUSE_ROOT_ID -> root_inode
+        // insert (FUSE_ROOT_ID -> root_inode) into KV engine
         meta.set_node_to_kv_engine(FUSE_ROOT_ID, root_inode).await?;
-        // 2. full_path -> FUSE_ROOT_ID
-        meta.set_inum_to_kv_engine(&full_path, FUSE_ROOT_ID).await?;
         Ok((meta, Some(server)))
     }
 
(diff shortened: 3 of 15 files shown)


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -665,94 +656,137 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+                    &ValueType::Node(old_parent_node.into_serial_node()),
+                );
+            }
+
+            let success = txn.commit().await?;
+
+            if success {
+                break Ok(());
+            }
+            debug!("exchange() found transaction commit failed when exchange, retrying...");
```
xiaguan, line 788: This loop potentially become an infinite loop in some cases. We should limit the retry times 🤔 ?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -1094,6 +1078,140 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
...
+                    parent,
+                    parent_node.get_name(),
+                ),
+            )
+        };
+
+        let old_parent_node = txn
+            .get(&KeyType::INum2Node(old_parent))
+            .await?
+            // return ENOENT instead of InconsistentFS, because this function may run multiple times
```
xiaguan, line 1146: The operation observes that the node doesn't exist; it should return ENOENT. This is similar to how some read-only operations don't require transaction protection.

src/async_fuse/memfs/s3_metadata.rs
Reto911: The commit operation fails, if the key is updated after gotten and before committed, right? Such cases may happen in a concurrent write scene, but would it become an infinite loop? It does seem to be like "starving", but I am sorry that I can't think of such a case in normal scene.

Besides, if we should limit the retry times, what type of error should it return when it hit the limit? Is it `TransactionRetryLimitExceededErr`?

src/async_fuse/memfs/s3_metadata.rs
Reto911: If the node exists, it would be modified in `rename_exchange_helper()`, so it's not a read-only operation but a read-modified-write operation, I think it does require transaction protection.

If it doesn't exist, the pre-check will fail, and `rename_exchange_helper()` will return immediately, and discards the txn. This should make no difference, I think🤔

xiaguan: What I mean is that you should delete this comment 🥵

src/async_fuse/memfs/s3_metadata.rs
xiaguan: The probability of this case occurring is very,very slow, but there always needs to be a limit for some such rare case 🥰

Reto911: Okey, thank you for your suggestion, it's a really significant lesson for me. And it's fixed now.

src/async_fuse/memfs/dir.rs
```diff
@@ -132,6 +132,12 @@ impl DirEntry {
         self.file_attr.read().kind
     }
 
+    /// Set name of the `DirEntry`.
+    #[allow(dead_code)]
```
xiaguan, line 136: Did you forget to remove this 🤔 ?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -54,6 +56,9 @@ const MY_TTL_SEC: u64 = 3600; // TODO: should be a long value, say 1 hour
 const MY_GENERATION: u64 = 1; // TODO: find a proper way to set generation
 /// S3 information string delimiter
 const S3_INFO_DELIMITER: char = ';';
+#[allow(dead_code)]
```
xiaguan, line 59: Same above

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -665,94 +659,132 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
         param: RenameParam,
     ) -> DatenLordResult<()> {
-        self.rename_exchange_local(context, &param).await?;
-        Ok(())
+        let old_parent = param.old_parent;
+        let old_name = param.old_name.as_str();
+        let new_parent = param.new_parent;
+        let new_name = param.new_name.as_str();
+        let flags = param.flags;
+        debug!(
```
xiaguan, line 691: Remove this log. The `#instrument` will print same.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -797,7 +829,7 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
         flags: u32,
     ) -> DatenLordResult<usize> {
         let data_len = data.len();
-        let (result, _, _) = {
+        let (result, _) = {
```
xiaguan, line 832: If we don't need this varaible anymore , we should do refactor here.

src/async_fuse/test/integration_tests.rs
```diff
@@ -296,6 +296,125 @@ fn test_rename_file_replace(mount_dir: &Path) -> anyhow::Result<()> {
     Ok(())
 }
 
+#[cfg(test)]
+#[cfg(feature = "abi-7-23")]
+fn test_rename_exchange(mount_dir: &Path) -> anyhow::Result<()> {
+    use nix::{fcntl::RenameFlags, sys::stat};
```
xiaguan, line 302: Use `std::fs` first. The version of  out `nix` is outdated , I'm not sure it is safe. Perhaps you could use ChatGpt to refactor this test 😆

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -135,6 +136,8 @@ pub enum KeyType {
     /// Node list
     /// The corresponding value type is ValueType::RawData
     FileNodeList(INum),
+    /// Just a string key for testing the KVEngine.
+    String(String),
```
rogercloud, line 140: If it's only for test, shall we add `#[test]`?

src/async_fuse/memfs/s3_metadata.rs
rogercloud: There're some functions with the similar issue above. We should take a look. Well maybe not this PR, it deserves a separate PR.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -1094,6 +1076,136 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
...
+                    parent_node.get_name(),
+                ),
+            )
+        };
+
+        let old_parent_node = txn
+            .get(&KeyType::INum2Node(old_parent))
+            .await?
+            .ok_or_else(|| {
+                anyhow::Error::new(Errno::ENOENT)
```
rogercloud, line 1145: Should we use anyhow error here? It's a typical error telling some resource doesn't exist. I think it's a "large" error deserves a datenlord error type.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -1094,6 +1076,136 @@ impl<S: S3BackEnd + Send + Sync + 'static> S3MetaData<S> {
...
+        };
+
+        let new_parent_node = if old_parent == new_parent {
+            None
+        } else {
+            let new_parent_node = txn
+                .get(&KeyType::INum2Node(new_parent))
+                .await?
+                .ok_or_else(|| {
+                    anyhow::Error::new(Errno::ENOENT)
```
rogercloud, line 1174: The same as above.

src/async_fuse/memfs/dir.rs
Reto911: No.

This method is called by `rename_exchange_helper`, but it's only available in 7.23. In the ABI version under 7.23, the method `set_entry_name` is not called, that's why I added `#[allow(dead_code)]` to it. And this method is not ABI specified, so I can't add `#[cfg(feature = "abi-7-23")]` to it. That's why I reserved `#[allow(dead_code)]`  here.

src/async_fuse/memfs/s3_metadata.rs
Reto911: No, and the reason is same as above.

The constant is used only by `rename_exchange_helper`, over 7.23. Maybe I can move this constant into the method, but the constant may be used in other methods in the future, as it is a common constant.

src/async_fuse/memfs/s3_metadata.rs
Reto911: Ok, I will refactor such functions when I finished the tasks at hands

src/async_fuse/test/integration_tests.rs
Reto911: Sorry, I can't. In fact, exchange rename, even the whole `renameat2` is a Linux-specific operation. Therefore, the rust standard library does not provide a function to do so. That's why I use `nix` here. Maybe I can use `std::fs::metadata` instead of `nix::sys::stat::stat`, however, `Metadata` in `std` doesn't implement the trait `Eq`, so it's hard to check if two files are actually exchanged.

`nix` may be the safest and most convenient way to test this feature now (compares to `libc`), and I can only try to rename some variables to make the codes clearer.

BTW, I don't have an openAI account, and my Microsoft account is banned from accessing Bing-chat😥

src/async_fuse/memfs/s3_metadata.rs
Reto911: It does deserve a DatenLord error type, but we cannot just return a value of other error types here now.

The codes here are imitating the widely used function `build_error_result_from_errno`, which returns an anyhow error:

```rust
pub fn build_error_result_from_errno<T>(error_code: Errno, err_msg: String) -> DatenLordResult<T> {
    Err(DatenLordError::from(
        anyhow::Error::new(error_code).context(err_msg),
    ))
}
```

The callers (in fact, they are methods of `MemFs` as `FileSystem` implementation) of those methods in `Metadata` will get a returned `DatenLordError`, and call `reply.error()` to send it to the kernel, if it is an `Err`.  The `error()` is defined as below:

```rust
pub async fn error(self, err: DatenLordError) -> nix::Result<usize> {
    self.reply.send_error(err).await
}
```
The `self.reply` is an instance of `ReplyRaw`, and `ReplyRaw::send_error()` is defined as:

```rust
async fn send_error(self, err: DatenLordError) -> nix::Result<usize> {
    match err {
        DatenLordError::InternalErr { source, context } => {
            let error_code = if let Some(nix_err) =
                source.root_cause().downcast_ref::<nix::Error>()
            {
                if *nix_err == nix::errno::Errno::UnknownErrno {
                    panic!(
                        "should not send nix::errno::Errno::UnknownErrno to FUSE kernel, \
                                the error is: {} ,context is : {:?}",
                        crate::common::util::format_anyhow_error(&source),
                        context,
                    );
                } else {
                    nix_err
                }
            } else {
                panic!(
                        "should not send non-nix error to FUSE kernel, the error is: {},context is : {:?}",
                        crate::common::util::format_anyhow_error(&source),context,
                    );
            };
            self.send_error_code(*error_code).await
        }
        err => {
            panic!("should not send non-internal error to FUSE kernel ,the error is : {err}",);
        }
    }
}
```

That's the problem! The `err` parameter must be a `DatenLordError::InternalErr`, with an `anyhow::Error` inside, and must be able to be downcast to an `nix::Errno`, otherwise the application will panic.

That's why the function `build_error_result_from_errno` and I return an anyhow error but not a seperated error. To solve such problems, we may need to refactor the mechanism of error handling of DatenLord.

src/async_fuse/memfs/dir.rs
Reto911: Well, I removed this method, and impl it in another way.

src/async_fuse/memfs/kv_engine/mod.rs
Reto911: Yes, I am sorry for my cursoriness.

scripts/setup/start_local_node.sh
```diff
@@ -9,6 +9,8 @@ export RUST_BACKTRACE=1
 export ETCD_END_POINT=127.0.0.1:2379
 export BIND_MOUNTER=`realpath $BIND_MOUNTER`
 
+BUILD_FLAGS=$1
```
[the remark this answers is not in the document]
Reto911: OK, I have fixed it up.

xiaguan (approved):

xiaguan:
Open an issue to document the `datenlord` error handling related problems and ideas.

PR #426, opened 2023-11-15
fix: Check ABI flags before 7.24 and add UT cases for FUSE requests

Reto911:
This PR mainly aims at check if the structures about FUSE protocol match their ABI flag marks (`#[cfg(feature = "abi-7-9")]` for example).

What the commits in this PR did are listed below:

1. Removed the support for MacOS from DatenLord.
2. Checked and fixed some ABI marks before 7.24.
3. Fix `SetAttr`, about `FATTR_ATIME_NOW`, `FATTR_MTIME_NOW` and `ctime` setting.
4. Implemented a crate `macro-utils` for generating test data, and added some UT cases for FUSE requests with it.


xiaguan (approved):
LGTM

xiaguan:
The following is merely an idea and has no direct relation to this PR. We have the capability to construct fuse requests. In light of this, is it feasible to view `datenlord` as a server operating under a Client/Server model to handle fuse requests? We could use automation to generate random requests. While it's expected that `datenlord` may return errors, encountering a `panic` should be avoided. For example, the `sqlsmith`  for databases. Again, this is just a conceptual idea 👓

Reto911:
> The following is merely an idea and has no direct relation to this PR. We have the capability to construct fuse requests. In light of this, is it feasible to view `datenlord` as a server operating under a Client/Server model to handle fuse requests? We could use automation to generate random requests. While it's expected that `datenlord` may return errors, encountering a `panic` should be avoided. For example, the `sqlsmith` for databases. Again, this is just a conceptual idea 👓

Nice idea!

Reto911:
Something wrong with ABI flags about some FUSE operation, such `Poll` is received at the version of 7.8, which raised a panic and made KIND-K8S-CSI-E2E test failed. I'm trying to fix such problems, so this PR is convrted to a draft now.

Reto911:
> Something wrong with ABI flags about some FUSE operation, such as that `Poll` is received at the version of 7.8 (however this feature should have been introduced in 7.11), which raised a panic (unknown FUSE opcode = 40) and made KIND-K8S-CSI-E2E test failed. I'm trying to fix such problems, so this PR is convrted to a draft now.

It's fixed up. Now the `Deserializer` will panic no more when an unknown operation found now, instead, it returns a `DeserializeError::UnknwonUnknownOpCode` and the DatenLord replies an `ENOSYS` to kernel.

xiaguan (dismissed):

macro-utils/src/generator.rs
```diff
@@ -0,0 +1,337 @@
...
+        $(
+            else if lookahead.peek(kw::$kws) {
+                $input.parse::<kw::$kws>()?;
+                $input.parse::<Token![:]>()?;
+                let $token_name = $input.parse::<$tokens>()?;
+                $blks
+            }
+        )*
+        else {
+            Err(lookahead.error())
```
rogercloud, line 136: Can we list all possible keywords in the error message? It will help debug.

macro-utils/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    // clippy::panic_in_result_fn,
+    clippy::separated_literal_suffix, // conflict with unseparated_literal_suffix
+    clippy::shadow_unrelated, //it’s a common pattern in Rust code
+    clippy::shadow_reuse, //it’s a common pattern in Rust code
+    clippy::shadow_same, //it’s a common pattern in Rust code
+    clippy::missing_errors_doc, // TODO: add error docs
+    clippy::exhaustive_structs,
+    clippy::exhaustive_enums,
+    clippy::missing_panics_doc, // TODO: add panic docs
+    clippy::panic_in_result_fn,
```
rogercloud, line 89: Please double check the allow lints. I think at least this one is not needed.

src/async_fuse/fuse/abi_marker.rs
```diff
@@ -110,16 +110,16 @@ mark_sized_types!(@kernel size_check: check_common,
...
-    // TODO: remove this after implement corresponding FUSE operations
-    // ---
-    FuseIoCtlIn,
-    FusePollIn,
-    FuseBatchForgetIn,
-    FuseForgetOne,
-    // ---
 );
 
+// TODO: remove this after implement corresponding FUSE operations
```
rogercloud, line 115: Why are they moved out?

src/async_fuse/fuse/de.rs
```diff
@@ -161,6 +170,7 @@ impl<'b> Deserializer<'b> {
     }
 
     /// Fetch remaining bytes and transmute to a slice of target instances
+    #[allow(dead_code)]
```
rogercloud, line 173: Is this function still useful? or are we going to use it later?

Reto911: Yes, this function is used on 7.16 to parse `FUSE_BATCH_FORGET`, but unused on 7.8, so I have to mark it with `#[allow(dead_code)]`.

There is a similar problem in #423 .

macro-utils/src/generator.rs
Reto911: `lookahead.error()` has done it, this function will record all keywords it peeked, and generate an error to list them.

By the way, you can see `macro-utils/tests/ui/generate/fails/*.rs` for test cases when the macro calling fails, and the `.stderr` files with the same names as `.rs` files include the error messages from compiler.

src/async_fuse/fuse/abi_marker.rs
Reto911: Because they are seperated into different ABI versions: `FuseIoCtlIn` and `FusePollIn` for 7.11, `FuseBatchForgetIn` and `FuseForgetOne` for 7.16. Now these structs are marked with correct ABI flags, and they need to be put in the correct macro call, so the TODO comment is moved out.

src/async_fuse/fuse/de.rs
rogercloud: Shall we mark it with `#[cfg(...)]` instead of "dead_code"?

Reto911: We can do it, this function is not such ABI related, but it's certainly not used before 7.16. And the function `fetch_slice` above is not used in any version, perhaps we can remove it.

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -679,47 +642,41 @@ impl ReplyDirectory {
     /// The kernel uses these value to request the next entries in further
     /// readdir calls
     pub fn add<T: AsRef<OsStr>>(&mut self, ino: u64, offset: i64, kind: SFlag, name: T) -> bool {
-        /// <https://doc.rust-lang.org/std/alloc/struct.Layout.html#method.padding_needed_for>
-        ///
-        /// <https://doc.rust-lang.org/src/core/alloc/layout.rs.html#226-250>
-        const fn round_up(len: usize, align: usize) -> usize {
-            len.wrapping_add(align).wrapping_sub(1) & !align.wrapping_sub(1)
-        }
+        use super::super::util;
```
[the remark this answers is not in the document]
Reto911: This is not such situation. Mod `util` is exactly defined at `super::super::util`.

src/async_fuse/fuse/de.rs
```diff
@@ -308,7 +292,9 @@ mod tests {
 
     #[test]
     #[allow(clippy::unwrap_used)]
+    #[cfg(feature = "abi-7-16")]
     fn fetch_all_as_slice() {
+        use super::DeserializeError;
```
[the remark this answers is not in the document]
Reto911: `DeserializeError` is defined at `super`, not imported in `super`, should I still make it `super::DeserializeError::NotEnough`?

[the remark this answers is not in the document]
Reto911: Sorry, I misunderstood you words😂

rogercloud (approved):

Issue #427, opened 2023-11-27
Transition to clap for Argument Parsing and Config Management


xiaguan:
Currently, our system uses hand-written argument parsing (`main.rs`), and as our system evolves, we will be adding more configuration parameters. Continuing with hand-written parsing not only adds extra work during updates but also becomes increasingly complex and difficult to maintain. We could use `clap` to create a separate config module.

- [x] Complete the `config` mod (using 'clap`)
- [x] Replace the hand-written parsing in `main.rs`

During implementation, it's important to consider the various configurations that may be added in the future.

xiaguan:
If you're interested in contributing to the project, I think this would be a great starting point.

PR #429, opened 2023-11-27
fix(fs): ignore check_perm when default_permission

xiaguan:
If fuse mount with '-o default_permissions' , then we should not check permission, the kernel will do this.
If we mount with that arg and still check permission by ourself ,following test case will failed (expect to success).
```rust
use std::fs::File;
use std::io;
use std::os::unix::fs::OpenOptionsExt;

fn main() -> io::Result<()> {
    let filename = "test";

    // Step 1: Create the file with mode 444
    let file = File::options()
        .create_new(true)
        .read(true)
        .write(true)
        .mode(0o444)
        .open(filename);

    match file {
        Ok(_) => println!("Everything is ok"),
        Err(e) => eprintln!("Error creating file: {}", e),
    }

    Ok(())
}
```
We don't need to perform additional permission checks when mount with default_permission.


xiaguan:
Using a constant bool value is not the best practice. After waiting for the `config` mod to be finished #427 , we will conduct permission checks by detecting whether `default_permission` is enabled.

Reto911:
Some functions checked the permission without calling `check_perm`, shall we double check them?

For example, in `S3Node::setattr_precheck()`:

```rust
//  owner is root check the user_id
if cur_attr.uid == 0 && user_id != 0 {
    return build_error_result_from_errno(
        Errno::EPERM,
        "setattr() cannot change atime".to_owned(),
    );
}
self.attr.read().check_perm(user_id, group_id, 2)?;
if user_id != cur_attr.uid {
    return build_error_result_from_errno(
        Errno::EACCES,
        "setattr() cannot change atime".to_owned(),
    );
}
```

This checked the root owned files. This may not be a big problem.

xiaguan:
> Some functions checked the permission without calling `check_perm`, shall we double check them?
> 
> For example, in `S3Node::setattr_precheck()`:
> 
> ```rust
> //  owner is root check the user_id
> if cur_attr.uid == 0 && user_id != 0 {
>     return build_error_result_from_errno(
>         Errno::EPERM,
>         "setattr() cannot change atime".to_owned(),
>     );
> }
> self.attr.read().check_perm(user_id, group_id, 2)?;
> if user_id != cur_attr.uid {
>     return build_error_result_from_errno(
>         Errno::EACCES,
>         "setattr() cannot change atime".to_owned(),
>     );
> }
> ```
> 
> This checked the root owned files. This may not be a big problem, but we can fix it in the future.

Yes, and `sticky bit` related check is same.  Kernel will check them.
I can fix some.

src/async_fuse/test/integration_tests.rs
```diff
@@ -1,3 +1,7 @@
+use std::fs::File;
+use std::io;
+use std::io::Write;
+use std::os::unix::fs::OpenOptionsExt;
 use std::path::Path;
 use std::{fs, iter};
```
rogercloud, line 6: Please format it and consider the ordering.

src/async_fuse/test/integration_tests.rs
```diff
@@ -634,13 +638,34 @@ fn test_delete_file(mount_dir: &Path) -> anyhow::Result<()> {
...
         Ok(_) => panic!("File deletion failed"),
-        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => {} // expected this error
+        Err(ref e) if e.kind() == io::ErrorKind::NotFound => {} // expected this error
         Err(e) => return Err(e.into()),
     }
 
     Ok(())
 }
 
+#[cfg(test)]
```
rogercloud, line 648: Please add a failed case to ensure that permission check really works.

Reto911 (approved):
LGTM

rogercloud:
> > Some functions checked the permission without calling `check_perm`, shall we double check them?
> > For example, in `S3Node::setattr_precheck()`:
> > ```rust
> > //  owner is root check the user_id
> > if cur_attr.uid == 0 && user_id != 0 {
> >     return build_error_result_from_errno(
> >         Errno::EPERM,
> >         "setattr() cannot change atime".to_owned(),
> >     );
> > }
> > self.attr.read().check_perm(user_id, group_id, 2)?;
> > if user_id != cur_attr.uid {
> >     return build_error_result_from_errno(
> >         Errno::EACCES,
> >         "setattr() cannot change atime".to_owned(),
> >     );
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This checked the root owned files. This may not be a big problem, but we can fix it in the future.
> 
> Yes, the situation with the `sticky bit` related checks is similar. The kernel will handle these checks. I can refactor some of the code accordingly.

Will kernel check the sticky bit for us?

rogercloud (approved):

xiaguan:
> > > Some functions checked the permission without calling `check_perm`, shall we double check them?
> > > For example, in `S3Node::setattr_precheck()`:
> > > ```rust
> > > //  owner is root check the user_id
> > > if cur_attr.uid == 0 && user_id != 0 {
> > >     return build_error_result_from_errno(
> > >         Errno::EPERM,
> > >         "setattr() cannot change atime".to_owned(),
> > >     );
> > > }
> > > self.attr.read().check_perm(user_id, group_id, 2)?;
> > > if user_id != cur_attr.uid {
> > >     return build_error_result_from_errno(
> > >         Errno::EACCES,
> > >         "setattr() cannot change atime".to_owned(),
> > >     );
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This checked the root owned files. This may not be a big problem, but we can fix it in the future.
> > 
> > 
> > Yes, the situation with the `sticky bit` related checks is similar. The kernel will handle these checks. I can refactor some of the code accordingly.
> 
> Will kernel check the sticky bit for us?

Kernel will check it.

PR #431, opened 2023-11-29
feat: Introduce `EvictPolicy`, `Storage` and related utilities

Reto911:
This PR aims at implementing `StorageManager` as a new storage managing mechanism, introducing some trait and utilities for it, and implementing the `InMemoryCache`, which is plugged into the new mechanism, as a replacement of the old `GlobalCache`.

All targets in this PR are listed below:

- [x] Introduce `EvictPolicy` trait and implement `LruPolicy`
- [x] Implement `StorageManager`
- [x] Implement `InMemoryCache`
- [x] Test `InMemoryCache`
- [x] Plug `InMemoryCache` into `StorageManager` and test the manager
- [x] Find the best way to write a block

When #432 is merged, clippy lints should be fixed:

- [x] Fix clippy lints.

Because the patches are too large, thus the `InMemoryCache` and `StorageManager` have been moved to #442.

### DIFF
--- a/src/async_fuse/memfs/cache/block.rs
+++ b/src/async_fuse/memfs/cache/block.rs
@@ -1,0 +1,320 @@
+//! Utilities for blocks.
+
+use std::{fmt::Formatter, sync::Arc};
+
+use aligned_utils::bytes::AlignedBytes;
+use clippy_utilities::OverflowArithmetic;
+use nix::sys::uio::IoVec;
+
+use crate::async_fuse::fuse::fuse_reply::{AsIoVec, CouldBeAsIoVecList};
+use crate::async_fuse::fuse::protocol::INum;
+
+/// Page Size
+const PAGE_SIZE: usize = 4096;
+
+/// A common coordinate to locate a block.
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct BlockCoordinate(pub INum, pub usize);
+
+/// Make a partial slice of `u8` slice for debugging.
+fn make_partial_u8_slice(src: &[u8], start: usize, len: usize) -> String {
+    let end = start.overflow_add(len).min(src.len());
+    let show_length = end.overflow_sub(start);
+    let slice = src.get(start..end).unwrap_or(&[]);
+
+    let mut result = String::from("[");
+
+    for (i, &ch) in slice.iter().enumerate() {
+        result.push_str(format!("{ch}").as_str());
+        if i != show_length.overflow_sub(1) {
+            result.push_str(", ");
+        }
+    }
+
+    if end != src.len() {
+        result.push_str(", ...");
+    }
+
+    result.push(']');
+
+    result
+}
+
+/// The minimum unit of data in the storage layers.
+#[derive(Clone)]
+pub struct Block {
+    /// The underlying data of a block. Shared with `Arc`.
+    inner: Arc<AlignedBytes>,
+    /// A flag that if this block is dirty.
+    dirty: bool,
+}
+
+impl std::fmt::Debug for Block {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let inner_slice = make_partial_u8_slice(self.inner.as_ref(), 0, 8);
+
+        let result = format!("Block {{ inner: {inner_slice}, dirty: {} }}", self.dirty());
+
+        write!(f, "{result}")
+    }
+}
+
+impl Block {
+    /// Create a block with `capacity`, which usually equals the `block_size` of storage manager.
+    #[must_use]
+    pub fn new(capacity: usize) -> Self {
+        Block {
+            inner: Arc::new(AlignedBytes::new_zeroed(capacity, PAGE_SIZE)),
+            dirty: false,
+        }
+    }
+
+    /// Returns the length of the block, which usually equals the `block_size` of storage manager.
+    #[must_use]
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    /// Check if the block is with size of 0.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    /// Get a mutable slice of the underlying data, copy them if there are other blocks hold the same data with `Arc`.
+    /// See also [`Arc::make_mut`](fn@std::sync::Arc::make_mut).
+    pub fn make_mut(&mut self) -> &mut [u8] {
+        Arc::make_mut(&mut self.inner).as_mut()
+    }
+
+    /// Checks if the block is dirty.
+    #[must_use]
+    pub fn dirty(&self) -> bool {
+        self.dirty
+    }
+
+    /// Sets the block to be dirty.
+    pub fn set_dirty(&mut self) {
+        self.dirty = true;
+    }
+}
+
+/// A wrapper for `IoBlock`, which is used for I/O operations
+#[derive(Clone)]
+pub struct IoBlock {
+    /// The inner `Block` that contains data
+    inner: Block,
+    /// The offset for this `Block`
+    offset: usize,
+    /// The end offset for this `Block`
+    end: usize,
+}
+
+impl std::fmt::Debug for IoBlock {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let inner_slice = make_partial_u8_slice(self.as_slice(), 0, 8);
+
+        let result = format!(
+            "IoBlock {{ inner: {inner_slice}, offset: {}, end: {}, dirty: {} }}",
+            self.offset,
+            self.end,
+            self.dirty()
+        );
+
+        write!(f, "{result}")
+    }
+}
+
+impl IoBlock {
+    /// The constructor of `IoBlock`
+    #[must_use]
+    pub fn new(inner: Block, offset: usize, end: usize) -> Self {
+        debug_assert!(offset <= end);
+        debug_assert!(
+            end <= inner.len(),
+            "The end {end} of slice is out of range of the inner block."
+        );
+        Self { inner, offset, end }
+    }
+
+    /// The inner block
+    #[must_use]
+    pub fn block(&self) -> &Block {
+        &self.inner
+    }
+
+    /// The offset of valid bytes of the inner block
+    #[must_use]
+    pub const fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// The end offset of valid bytes of the inner block
+    #[must_use]
+    pub const fn end(&self) -> usize {
+        self.end
+    }
+
+    /// Turn `IoBlock` into slice
+    pub(crate) fn as_slice(&self) -> &[u8] {
+        self.inner
+            .inner
+            .get(self.offset..self.end)
+            .unwrap_or_else(|| {
+                unreachable!(
+                    "`{}..{}` is checked not to be out of range of inner block.",
+                    self.offset, self.end,
+                )
+            })
+    }
+
+    /// Checks if the inner block is dirty.
+    #[must_use]
+    pub fn dirty(&self) -> bool {
+        self.inner.dirty
+    }
+
+    /// Sets the inner block to be dirty.
+    pub fn set_dirty(&mut self) {
+        self.inner.dirty = true;
+    }
+}
+
+#[cfg(test)]
+impl From<Block> for IoBlock {
+    fn from(block: Block) -> Self {
+        let len = block.len();
+        IoBlock::new(block, 0, len)
+    }
+}
+
+impl CouldBeAsIoVecList for IoBlock {}
+
+impl AsIoVec for IoBlock {
+    fn as_io_vec(&self) -> IoVec<&[u8]> {
+        IoVec::from_slice(self.as_slice())
+    }
+
+    fn can_convert(&self) -> bool {
+        true
+    }
+
+    fn len(&self) -> usize {
+        self.end.overflow_sub(self.offset)
+    }
+
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use clippy_utilities::OverflowArithmetic;
+
+    use crate::async_fuse::fuse::fuse_reply::AsIoVec;
+
+    use super::{Arc, Block, IoBlock};
+
+    const BLOCK_SIZE_IN_BYTES: usize = 8;
+    const BLOCK_CONTENT: &[u8; BLOCK_SIZE_IN_BYTES] = b"foo bar ";
+
+    #[test]
+    fn test_block() {
+        let mut block = Block::new(BLOCK_SIZE_IN_BYTES);
+        assert_eq!(block.len(), BLOCK_SIZE_IN_BYTES);
+        assert!(!block.is_empty());
+
+        block.make_mut().copy_from_slice(BLOCK_CONTENT);
+        assert_eq!(block.inner.get(..), Some(BLOCK_CONTENT.as_slice()));
+
+        let _another_block = block.clone();
+        assert_eq!(Arc::strong_count(&block.inner), 2);
+
+        assert_eq!(
+            format!("{block:?}"),
+            "Block { inner: [102, 111, 111, 32, 98, 97, 114, 32], dirty: false }"
+        );
+    }
+
+    #[test]
+    fn test_small_block() {
+        let mut block = Block::new(4);
+        assert_eq!(block.len(), 4);
+
+        block.make_mut().copy_from_slice(&BLOCK_CONTENT[..4]);
+        assert_eq!(block.inner.get(..), Some(&BLOCK_CONTENT[..4]));
+
+        assert_eq!(
+            format!("{block:?}"),
+            "Block { inner: [102, 111, 111, 32], dirty: false }"
+        );
+    }
+
+    #[test]
+    fn test_large_block() {
+        let mut block = Block::new(BLOCK_SIZE_IN_BYTES.overflow_mul(2));
+        assert_eq!(block.len(), BLOCK_SIZE_IN_BYTES.overflow_mul(2));
+
+        block
+            .make_mut()
+            .copy_from_slice(BLOCK_CONTENT.repeat(2).as_slice());
+
+        assert_eq!(
+            format!("{block:?}"),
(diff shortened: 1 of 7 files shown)


xiaguan:
1. In `mod.rs` ,we only handle module declarations and related concerns. We avoid placing other logic in it and strive to split the code as mucch as possible for better organization.
2. The methods `StoratgeManager::load()` and `StorageManger::store` are too lengthy and need to be refactored for imporved readability and maintainability.

Reto911:
The block overwriting policy is CoW now, and there are 3 solutions:
1. Use `remove`-`flush`-`repin`-`flush` to ensure that a block is removed from the hash map, in order to try best to reduce copy during writing. But there may be extra cost.
2. Just `make_mut` on `Arc` in `Block`, which will certainly copy the data. This avoids the extra cost of crossbeam GC, but introduces the copy cost.
3. Use a locked `HashMap` in `std` as the file-level cache, which can certainly avoid the copy cost and GC cost, but reduces the concurrency.
4. In fact, there is the fourth solution: use a block-level lock (`RwLock<Block>`). If this is possible, this must be faster than the third.

We should run benchmarks to test the solutions above, and find the best way for block overwriting.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,416 @@
...
+    P: EvictPolicy<BlockCoordinate> + Send + Sync,
+    S: Storage + Send + Sync,
+{
+    async fn load(&self, ino: INum, block_id: usize) -> Option<Block> {
+        let block_in_cache = self.get_block_from_cache(ino, block_id);
+
+        match block_in_cache {
+            None => {
+                // If a fetching from backend still fails, just return a `None`.
+                let block_from_backend = self.backend.load(ino, block_id).await?;
```
Reto911, line 135: Fetching an inexisting block should return a new block filled with zeros.

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -0,0 +1,533 @@
...
+    ) -> Vec<IoBlock> {
+        // Check if the cache is valid.
+        let invalid = {
+            let guard = pin();
+            let cache_mtime = self.mtimes.get(&ino, &guard);
+            cache_mtime != Some(&mtime)
+        };
+
+        if invalid {
+            self.storage.invalidate(ino).await;
```
xiaguan, line 180: It should be an async task. We need to create an issue to document the design of asynchronous tasks for the storage module. We can do this later.

xiaguan: Or a flag called `valid` to reuse memory space? We could think it later. Create an issue about reusing memory.

xiaguan:
Finally, here are some organazition suggestions.
First, this module should be moved to src/storage, as it's no longer just a simple cache-related module.
Second, we need to add robust integration testing (IT) to this storage module, which can be done later. Casue this mod is **crucial** to us.

xiaguan:
I am surprised to see that you have used a lot of functional programming tricks 😄 👍

Reto911:
> I am surprised to see that you have used a lot of functional programming tricks 😄 👍

It is one of my loves😏

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,416 @@
...
+    #[tokio::test]
+    #[should_panic(expected = "out of range")]
+    async fn test_write_out_of_range() {
+        let policy = LruPolicy::<BlockCoordinate>::new(CACHE_CAPACITY_IN_BLOCKS);
+        let backend = Arc::new(MemoryStorage::new(BLOCK_SIZE_IN_BYTES));
+        let cache = InMemoryCache::new(policy, backend, BLOCK_SIZE_IN_BYTES);
+
+        let io_block = IoBlock::new(Block::new(16), 0, 16);
+        cache.store(0, 0, io_block).await;
+    }
```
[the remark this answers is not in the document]
xiaguan: Yes. Last time, my disk cache had a deadlock issue. 
Would it be better to include the concurrency tests in the integration testing (IT) of the entire module?

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,176 @@
...
+                    "`{}..{}` is out of range of {}.",
+                    self.offset,
+                    self.end,
+                    self.inner.len()
+                )
+            })
+    }
+}
+
+impl From<Block> for IoBlock {
```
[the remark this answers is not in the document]
Reto911: In fact it's mainly used in tests just for convenience. Or perhaps I should mark it with `#[cfg(test)]`?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,416 @@
...
+        self.backend.remove(ino).await;
+    }
+
+    async fn invalidate(&self, ino: INum) {
+        self.map.remove(&ino);
+        self.backend.invalidate(ino).await;
+    }
+
+    async fn flush(&self, ino: INum) {
+        self.backend.flush(ino).await;
```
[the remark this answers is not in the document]
Reto911: Yes, and I will add dirty-flag for block structures later.

src/async_fuse/memfs/cache/block.rs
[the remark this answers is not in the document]
xiaguan: Yes

src/async_fuse/memfs/cache/storage_manager.rs
Reto911: > It should be an async task. We need to create an issue to document the design of asynchronous tasks for the storage module. We can do this later.

But we must wait for the invalidation before we fetch a block from the storage, or we may success to load invalid blocks.

src/async_fuse/memfs/cache/memory_cache.rs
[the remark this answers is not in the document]
Reto911: I just created #440 for IT, and I would add a simple UT for `StorageManager`.

src/async_fuse/memfs/cache/storage_manager.rs
xiaguan: We could add a flag called `valid` to Block (like the dirty bit). Invalid operation just set the bit to false. Then we could reuse the Block's memory.  We could consider this when you start to refactor the old system.

Reto911: > We could add a flag called `valid` to Block (like the dirty bit). Invalid operation just set the bit to false. Then we could reuse the Block's memory. We could consider this when you start to refactor the old system.

I think it's not good, because the invalidation is on file-level now. I would rather destroy a file-level HashMap to do the invalidation, than iterate it and mark all the blocks as invalid. Besides, the space of a `Block` is shared via an `Arc`, therefore we can reuse the sapce only if the block is not referenced outside of the cache.

xiaguan: Reuse memory is complex. Destorying a file-level hashmap is a good idea.

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,176 @@
...
+        let len = block.len();
+        IoBlock::new(block, 0, len)
+    }
+}
+
+impl CouldBeAsIoVecList for IoBlock {}
+
+impl AsIoVec for IoBlock {
+    fn as_io_vec(&self) -> IoVec<&[u8]> {
+        assert!(self.can_convert(), "Please check before convert to IoVec");
```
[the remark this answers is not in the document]
Reto911: There is no need to check it anymore, because `self.can_convert()` of `IoBlock` will always return `true` .

src/async_fuse/memfs/cache/policy/lru.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        lru.insert(key, ());
+        evicted.map(|(k, ())| k)
+    }
+
+    fn touch(&self, key: &K) {
+        let _: Option<&()> = self.inner.lock().get(key);
+    }
+
+    fn remove(&self, key: &K) {
```
rogercloud, line 45: Shall we provide a function to get the removed entry?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,439 @@
...
+use lockfree_cuckoohash::{pin, LockFreeCuckooHash as HashMap};
+use parking_lot::RwLock;
+use std::collections::HashMap as StdHashMap;
+
+use super::policy::EvictPolicy;
+use super::{Block, BlockCoordinate, IoBlock, Storage};
+use crate::async_fuse::fuse::protocol::INum;
+
+/// Merge the content from `src` to `dst`.
+fn merge_two_blocks(dst: &mut Block, src: &IoBlock) {
```
rogercloud, line 13: We usually put src in front of dst.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,439 @@
...
+        let guard = pin();
+        let block = self
+            .map
+            .get(&ino, &guard)
+            .and_then(|file_cache| file_cache.read().get(&block_id).cloned());
+        block
+    }
+
+    /// Return true if success
+    fn merge_block_in_place(&self, ino: INum, block_id: usize, to_merge: &IoBlock) -> bool {
```
rogercloud, line 63: Would it be better if we rename `to_merge` to `src`?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,439 @@
...
+where
+    P: EvictPolicy<BlockCoordinate> + Send + Sync,
+    S: Storage + Send + Sync,
+{
+    /// Loads a block from the cache, or loads it from the backend.
+    ///
+    /// If backend returns a `None`, this will create a new block with zeros filled,
+    /// inserts it into the cache, and returns it to caller. That means, this method
+    /// will never return `None`.
+    async fn load(&self, ino: INum, block_id: usize) -> Option<Block> {
```
rogercloud, line 135: I think the main logic of load and store is the same for both Memory and Disk.

For read:
1. read local cache.
    1. if local cache is missing, load the backend.
2. if there's a load function in the previous step, we may get a evicted block, handle it properly.
3. return actul data to the caller.

According to the previous steps, we can make a default implementation for the load function, and move the `load_from_local_cache` into the trait.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,439 @@
...
+        // The `InMemoryCache` cannot iterate its `HashMap`, thus it cannot `flush_all` on all the file cache.
+        // As it runs with a "write through" policy, the backend takes the responsibility to do the actual `flush_all`.
+        self.backend.flush_all().await;
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use std::sync::Arc;
```
rogercloud, line 231: Please make a large test function into small ones. It's easier to debug, and has better performance on power servers.

src/async_fuse/memfs/cache/mock.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use async_trait::async_trait;
+use parking_lot::Mutex;
+
+use crate::async_fuse::fuse::protocol::INum;
+
+use super::{Block, IoBlock, Storage};
+
+/// A "persistent" storage layer in memory.
+#[derive(Default)]
```
rogercloud, line 13: If these data structures are used in test only. Please mark it with `#[test]`.

Reto911: I'm not sure if it is only used in test, we may use it as a backend when running local node, until the solution of backend interface is stable.

Reto911:
I may divide this PR into 2, because this PR is too large now.

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -0,0 +1,775 @@
...
+        let to_block = self.offset_to_block_id(to.overflow_sub(1)).overflow_add(1);
+
+        self.storage.truncate(ino, from_block, to_block).await;
+
+        // fill the rest bytes with zeros.
+        let fill_end = to_block.overflow_mul(self.block_size);
+        let fill_start = to;
+        let fill_len = fill_end.overflow_sub(fill_start);
+        let fill_content = vec![0_u8; fill_len];
+        self.store(ino, fill_start, &fill_content, mtime).await
```
Reto911, line 323: We can do this in `Storage::truncate`, in order to handle the truncating in a single function call.

src/async_fuse/memfs/cache/mod.rs
```diff
@@ -1,5 +1,39 @@
...
+pub const KB_SIZE: usize = 1024;
+
+/// The number of bytes in one MiB.
+pub const MB_SIZE: usize = 1024 * KB_SIZE;
+
+/// The number of bytes in one GiB.
+pub const GB_SIZE: usize = 1024 * MB_SIZE;
+
+/// The size of a block.
+pub const BLOCK_SIZE_IN_BYTES: usize = 512 * KB_SIZE;
```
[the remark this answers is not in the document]
Reto911: We can list all configurable items in a document. In this case, variables such as `block size` and `memory cache capacity` are configurable.

src/async_fuse/memfs/cache/storage_manager.rs
Reto911: There is no need to handle the truncating in a single function call, although this causes multiple calls on disk cache.

There should not be one-to-one correspondence between function call and disk access in the disk cache, the cache has the responsibility to optimize the IO operation.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: The `InMemoryCache` implementation has been moved to #442.

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    /// Get a mutable slice of the underlying data, copy them if there are other blocks hold the same data with `Arc`.
+    /// See also [`Arc::make_mut`](fn@std::sync::Arc::make_mut).
+    pub fn make_mut(&mut self) -> &mut [u8] {
+        Arc::make_mut(&mut self.inner).as_mut()
+    }
+
+    /// Checks if the block is dirty.
+    #[must_use]
+    pub fn dirty(&self) -> bool {
```
rogercloud, line 92: `is_empty` and `is_dirty`? or `empty` and `dirty`?

src/async_fuse/memfs/cache/storage.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// `[0, to_block)`.
+    async fn truncate(&self, ino: INum, from_block: usize, to_block: usize, fill_start: usize);
+
+    // Provided methods
+
+    /// Load a block from the storage.
+    async fn load(&self, ino: INum, block_id: usize) -> Option<Block> {
+        let block_in_cache = self.load_from_self(ino, block_id).await;
+
+        if block_in_cache.is_some() {
```
rogercloud, line 62: We can merge these two lines into `if let`.

src/async_fuse/memfs/cache/block.rs
Reto911: The `dirty` method is a getter of a private field, and it's a convention to name a getter in the same name as the field it gets in Rust.

The `is_empty` method is not a getter, and in fact, it's implemented to satisfy the clippy lints `clippy::len_without_is_empty`.

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,306 @@
...
+const PAGE_SIZE: usize = 4096;
+
+/// A common coordinate to locate a block.
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct BlockCoordinate(pub INum, pub usize);
+
+/// Make a partial slice of `u8` slice for debugging.
+fn make_partial_u8_slice(src: &[u8], start: usize, len: usize) -> String {
+    let end = start.overflow_add(len).min(src.len());
+    let show_length = end.overflow_sub(start);
```
xiaguan, line 22: By the way, the `overflow_add` and `overflow_sub` is hard to read.
```
let end = min(src.len(), start + len);
```
It's more easy to read.

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,306 @@
...
+impl From<Block> for IoBlock {
+    fn from(block: Block) -> Self {
+        let len = block.len();
+        IoBlock::new(block, 0, len)
+    }
+}
+
+impl CouldBeAsIoVecList for IoBlock {}
+
+impl AsIoVec for IoBlock {
```
xiaguan, line 193: Does  the trait `AsIoVec` make sense now?
I am not sure about its usage is outdated.

src/async_fuse/memfs/cache/block.rs
```diff
@@ -0,0 +1,306 @@
...
+    use crate::async_fuse::fuse::fuse_reply::AsIoVec;
+
+    use super::{Arc, Block, IoBlock};
+
+    const BLOCK_SIZE_IN_BYTES: usize = 8;
+    const BLOCK_CONTENT: &[u8; BLOCK_SIZE_IN_BYTES] = b"foo bar ";
+
+    #[test]
+    fn test_block() {
+        let mut block = Block::new(BLOCK_SIZE_IN_BYTES);
```
xiaguan, line 224: We need to test a small block which size is less than 8.

src/async_fuse/memfs/cache/block.rs
Reto911: Just using operatoer `+` will violate the clippy inspection rules.

src/async_fuse/memfs/cache/block.rs
Reto911: I think it's OK and there is no need to change it for now.

src/async_fuse/memfs/cache/block.rs
Reto911: Are you meaning to test the Debug information of `Block` that shorter than 8?

rogercloud (approved):

xiaguan (approved):

PR #432, opened 2023-12-01
chore: bump Rust version to 1.74.0

xiaguan:
When I wanted to use `clap` version 4 , I found that it requires Rust version >= `1.70.0`. To prevent similar situations in the future, I upgraded to the latest stable version (1.74.0). The upgrade itself will introduce an upgrade to clippy, bringing in new standards.

Clippy warnings mainly occur in the following modules:

1. `fuse`: Issues with byte order calling APIs in test.
2. `etcd_delegate` and `dist`: These need to be thoroughly cleaned up as we no longer require them. We should use `kv_engine` instead.
3. `csi`: This could be maintained as a separate crate or repository.

### DIFF
--- a/macro-utils/src/lib.rs
+++ b/macro-utils/src/lib.rs
@@ -79,7 +79,6 @@
     clippy::shadow_reuse, //it’s a common pattern in Rust code
     clippy::shadow_same, //it’s a common pattern in Rust code
     clippy::arithmetic_side_effects,  // An error is raised when calling `quote::quote!`
-    clippy::arithmetic_side_effects  // As same as above
 )]
 
 use proc_macro::TokenStream;
--- a/macro-utils/src/lib.rs
+++ b/macro-utils/src/lib.rs
@@ -78,7 +78,7 @@
     clippy::shadow_unrelated, //it’s a common pattern in Rust code
     clippy::shadow_reuse, //it’s a common pattern in Rust code
     clippy::shadow_same, //it’s a common pattern in Rust code
-    clippy::integer_arithmetic,  // An error is raised when calling `quote::quote!`
+    clippy::arithmetic_side_effects,  // An error is raised when calling `quote::quote!`
     clippy::arithmetic_side_effects  // As same as above
 )]
 
--- a/src/async_fuse/fuse/fuse_reply.rs
+++ b/src/async_fuse/fuse/fuse_reply.rs
@@ -24,7 +24,7 @@ use super::protocol::{
 #[cfg(feature = "abi-7-18")]
 use super::protocol::{FuseNotifyCode::FUSE_NOTIFY_DELETE, FuseNotifyDeleteOut};
 
-/// This trait describes a type that can be converted to Vec<IoVec<&[u8]>>
+/// This trait describes a type that can be converted to Vec<`IoVec`<&[u8]>>
 pub trait AsIoVecList {
     /// Convert the type to a Vec of `IoVec`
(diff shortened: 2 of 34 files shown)


xiaguan:
Related issues: #434  #433

macro-utils/src/lib.rs
```diff
@@ -78,7 +78,7 @@
     clippy::shadow_unrelated, //it’s a common pattern in Rust code
     clippy::shadow_reuse, //it’s a common pattern in Rust code
     clippy::shadow_same, //it’s a common pattern in Rust code
-    clippy::integer_arithmetic,  // An error is raised when calling `quote::quote!`
+    clippy::arithmetic_side_effects,  // An error is raised when calling `quote::quote!`
     clippy::arithmetic_side_effects  // As same as above
```
Reto911, line 82: Why do you amend the line 81 to `clippy::arithmetic_side_effects` while there is a `clippy::arithmetic_side_effects` below already?

src/common/error.rs
```diff
@@ -290,16 +290,16 @@ where
     where
         C: Into<String>,
     {
-        self.map_err(|e| e.into().add_context(ctx))
+        self.map_err(|err| err.into().add_context(ctx))
```
Reto911, line 293: Which clippy lint did this violate? Does it check an identifier with single letter?

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -24,7 +24,7 @@ use super::protocol::{
 #[cfg(feature = "abi-7-18")]
 use super::protocol::{FuseNotifyCode::FUSE_NOTIFY_DELETE, FuseNotifyDeleteOut};
 
-/// This trait describes a type that can be converted to Vec<IoVec<&[u8]>>
+/// This trait describes a type that can be converted to Vec<`IoVec`<&[u8]>>
```
Reto911, line 27: Which clippy lint did this violate? Shall we say `Vec<IoVec<&[u8]>>` here?

src/async_fuse/fuse/fuse_reply.rs
```diff
@@ -35,7 +35,7 @@ pub trait AsIoVecList {
 }
 
 /// Any type implement the `AsIoVec` trait, its Vec can be automatically for
-/// Vec<IoVec<&[u8]>>.
+/// Vec<`IoVec`<&[u8]>>.
```
Reto911, line 38: Same as above

src/async_fuse/memfs/serial.rs
```diff
@@ -79,7 +79,7 @@ pub enum SerialSFlag {
 }
 
 /// In order to derive Serialize and Deserialize,
-/// Replace the 'BTreeMap<String, `DirEntry`>' with 'HashMap<String,
+/// Replace the `BTreeMap`<String, `DirEntry`>' with `HashMap`<String,
```
Reto911, line 82: Same as above

src/async_fuse/fuse/fuse_reply.rs
xiaguan: Clippy did suggest me to do it.

Reto911: Sorry, I mean, add the backticks (\`) outside the whole `Vec<IoVec<&[u8]>>`, as it is a complete type.

xiaguan: I think current way is better. Cause `IoVec` is defined in this mod.

Reto911 (approved):

PR #438, opened 2023-12-05
chore: Move `GlobalCache` aside

Reto911:
In this PR, we moved the `GlobalCache` and related utilities to a seperated file, and re-import these stuffs in the mod file.

As the new storage manager is completed, the old `GlobalCache` will be removed formally.


src/async_fuse/memfs/cache/mod.rs
```diff
@@ -1,1022 +1,8 @@
 //! This is the cache implementation for the memfs
 
-use std::fmt::{Debug, Error, Formatter};
-use std::hash::{Hash, Hasher};
-use std::ops::Deref;
-use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+// TODO: Remove this after the storage is ready for product env.
+#![allow(dead_code)]
```
xiaguan, line 4: In this pr, I think `global_cache` is not dead code?

Reto911: Yes, this is kind of my negligence. However, several methods in `global_cache.rs` are also marked with `#[allow(dead_code)]`, so I think it's ok to reserve it as this commit is a temporary change.

If you think the `allow` attribute here is not good, I will move this line to #431 . Anyway, when we plug the new storage manager into our project (in a PR after #431 ), this `allow` must be removed finally.

xiaguan (dismissed):

rogercloud (dismissed):

Issue #425, opened 2023-11-14
Refactor the mechanism of error handling


Reto911:
The mechanism of error handling is not good now.

DatenLord use `DatenLordError` mainly to transmit errors, which has 26 variants now:

https://github.com/datenlord/datenlord/blob/83f3eb398592766517e40aaa52758a4b538846be/src/common/error.rs#L28-L268

It's too clumsy, and with this project being developed, there may be more variant joining into the enum. Complexity is not the only problem, in fact, there are something inconsistent in the mechanism.

There is a widely used helper function to build error from `nix::Errno`:

https://github.com/datenlord/datenlord/blob/83f3eb398592766517e40aaa52758a4b538846be/src/async_fuse/util.rs#L27-L31

It returns a `DatenLordError::InternalError`, with an `anyhow::Error` inside.  A `DatenLordResult` (may be with a `DatenLordError`) will be passed from `Metadata` to methods in `MemFs`, and they will call `reply.error()` to send it to the kernel, if it is an `Err`.  The `error()` calls `ReplyRaw::send_error()` to send the error to the kernel:

https://github.com/datenlord/datenlord/blob/83f3eb398592766517e40aaa52758a4b538846be/src/async_fuse/fuse/fuse_reply.rs#L229-L258

The `err` parameter must be a `DatenLordError::InternalErr`, with an `anyhow::Error` inside, and must be able to be downcast to an `nix::Errno`, otherwise the application will panic. However, we do have a `DatenLordError::NixError` consisting of a `nix::Errno`, which will raise panic if we pass it to `ReplyRaw::send_error()`. It doesn't make sence. 

Besides, some error types, such as `S3Error`, are not included in `DatenLordError`, to returns them from a function that returns `DatenLordResult`, we should convert it into an `anyhow::Error` and wrap it into `DatenLordError::InternalError`. But we should try to reduce the usage of `anyhow::Error`.

Just fixing the function `ReplyRaw::send_error()` and `build_error_result_from_errno` or adding more error types as variants into `DatenLordError` is short-sighted, this cannot solve the essential issues. Therefore, we should try to refactor the mechanism for convenience to maintain the project.

This problem is not such urgency, and there is not a sophisticated solution now, however, we should take attention to it and try to solve it in the future, before it's too hard for the increasing complexity.

PR #442, opened 2023-12-11
feat: Implement the `StorageManager` and `InMemoryCache`

Reto911:
`StorageManager` and `InMemoryCache` are implemented in this PR, they are moved from #431.

### DIFF
--- a/src/async_fuse/memfs/cache/memory_cache.rs
+++ b/src/async_fuse/memfs/cache/memory_cache.rs
@@ -1,0 +1,628 @@
+//! The in-memory cache.
+
+use async_trait::async_trait;
+use clippy_utilities::OverflowArithmetic;
+use lockfree_cuckoohash::{pin, LockFreeCuckooHash as HashMap};
+use std::collections::HashMap as StdHashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+use super::policy::EvictPolicy;
+use super::{Block, BlockCoordinate, BlockId, Storage};
+use crate::async_fuse::fuse::protocol::INum;
+
+/// The file-level cache.
+type FileCache = Arc<RwLock<StdHashMap<BlockId, Block>>>;
+
+/// Merge the content from `src` to `dst`. This will set `dst` to be dirty.
+fn merge_two_blocks(src: &Block, dst: &mut Block) {
+    dst.set_dirty();
+    dst.update(src);
+}
+
+/// The in-memory cache, implemented with lockfree hashmaps.
+#[derive(Debug)]
+pub struct InMemoryCache<P, S> {
+    /// The inner map where the cached blocks stored
+    map: HashMap<INum, FileCache>,
+    /// The evict policy
+    policy: P,
+    /// The backend storage
+    backend: S,
+    /// The block size
+    block_size: usize,
+}
+
+impl<P, S> InMemoryCache<P, S> {
+    /// The limit of retrying to insert a block into the cache.
+    const INSERT_RETRY_LIMMIT: usize = 10;
+
+    /// Create a new `InMemoryCache` with specified `policy`, `backend` and `block_size`.
+    pub fn new(policy: P, backend: S, block_size: usize) -> Self {
+        InMemoryCache {
+            map: HashMap::new(),
+            policy,
+            backend,
+            block_size,
+        }
+    }
+
+    /// Get the file-level cache `HashMap`.
+    fn get_file_cache(&self, ino: INum) -> Option<FileCache> {
+        let guard = pin();
+        self.map.get(&ino, &guard).cloned()
+    }
+
+    /// Get a block from the in-memory cache without fetch it from backend.
+    ///
+    /// Notice that this method will also not change the position of the block
+    /// in the policy.
+    async fn get_block_from_cache(&self, ino: INum, block_id: usize) -> Option<Block> {
+        let block = if let Some(file_cache) = self.get_file_cache(ino) {
+            file_cache.read().await.get(&block_id).cloned()
+        } else {
+            None
+        };
+
+        block
+    }
+
+    /// Update a block into cache in place.
+    /// Return the block if success (the destination existing in cache), otherwise returns `None`.
+    async fn update_block(&self, ino: INum, block_id: usize, src: &Block) -> Option<Block> {
+        let res = if let Some(file_cache) = self.get_file_cache(ino) {
+            let mut file_cache = file_cache.write().await;
+            if let Some(block) = file_cache.get_mut(&block_id) {
+                merge_two_blocks(src, block);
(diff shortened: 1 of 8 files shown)


src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,667 @@
...
+        ino: INum,
+        block_id: usize,
+        block: Block,
+    ) -> Option<(BlockCoordinate, Block)>
+    where
+        P: EvictPolicy<BlockCoordinate>,
+    {
+        let mut empty_file_cache = None;
+        let to_be_evicted = self.policy.put(BlockCoordinate(ino, block_id));
+        let guard = pin();
```
xiaguan, line 92: I believe there's gap here. Let's assume the cache capacity is 1.
If a thread1 in here put it's own key (1,1) ,and evict the (0,0).
When thread2 in , thread2 put it's own key (2,2), and evict the (1,1), but the thread1 haven't insert (1,1) into map.
So thread2 evict nothing.
Thread1 and thread2 then insert their own key into the map.
So we can left a (1,1) in map but will never be evicted, right ?
What I want to say is , the gap between policy and map is very dangerous.
I tryied to write a test to approve it, but I failed.
Maybe something is wrong?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,667 @@
...
+                Block::new(self.block_size)
+            })
+        };
+
+        merge_two_blocks(&block, &mut to_be_inserted);
+
+        self.write_and_evict(ino, block_id, to_be_inserted.clone())
+            .await;
+
+        self.backend
```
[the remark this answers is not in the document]
Reto911: We cannot simply call `load` funtion dispite of the difference of cache hit and cache miss. If we do so, we will be not able to merge the passed-in block to the cache in place, and the whole block must be copied.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,649 @@
...
+    /// If `backend` returns a `None`, this will create a new block with zeros filled,
+    /// inserts it into the `backend`, and returns it to caller. That means, this method
+    /// will never return `None`.
+    async fn load_from_backend(&self, ino: INum, block_id: usize) -> Option<Block> {
+        if let Some(from_backend) = self.backend.load(ino, block_id).await {
+            Some(from_backend)
+        } else {
+            // If a fetching from backend still fails, create a new block filled with zeros,
+            // then write it to cache and backend as stored.
+            let mut zeroed_block = Block::new(self.block_size);
```
xiaguan, line 186: `new` -> `new_zeroed`

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,649 @@
...
+
+    async fn cache_block_from_backend(&self, ino: INum, block_id: usize, block: Block) {
+        self.write_block_into_cache(ino, block_id, block).await;
+    }
+
+    async fn store(&self, ino: INum, block_id: usize, mut input: IoBlock) {
+        let start_offset = input.offset();
+        let end_offset = input.end();
+
+        assert!(end_offset <= self.block_size, "out of range");
```
xiaguan, line 210: use `debug_assert!` instead

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,649 @@
...
+            }
+            .await;
+
+            let success = self.policy.try_put(BlockCoordinate(ino, block_id));
+
+            if success {
+                break file_cache;
+            }
+            drop(file_cache);
+            self.evict().await;
```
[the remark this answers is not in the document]
Reto911: Why? Could you please explain how this works?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -0,0 +1,649 @@
...
+        .get_mut(start_offset..end_offset)
+        .unwrap_or_else(|| unreachable!("`src` is checked to be in range."))
+        .copy_from_slice(src.as_slice());
+}
+
+/// The in-memory cache, implemented with lockfree hashmaps.
+#[derive(Debug)]
+pub struct InMemoryCache<P, S> {
+    /// The inner map where the cached blocks stored
+    map: HashMap<INum, Arc<RwLock<StdHashMap<usize, Block>>>>,
```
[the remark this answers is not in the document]
Reto911: Do you mean to define `type BlockId = usize;` ?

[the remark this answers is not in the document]
xiaguan: Yes

src/async_fuse/memfs/cache/memory_cache.rs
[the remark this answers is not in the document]
xiaguan: Ignore this,sry.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: This is intentional but not for debugging, and will be replaced with an error returned in the future (the instant when error handling is introduced).

xiaguan: Link to the issue #425

xiaguan (approved):

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -0,0 +1,812 @@
...
+                .unwrap_or_else(|e| panic!("Fails when awaiting on storing: {e}"));
+        }
+    }
+
+    /// Convert `IoBlock`s from slice.
+    fn make_io_blocks_from_slice(
+        &self,
+        offset: usize,
+        data: &[u8],
+    ) -> impl Iterator<Item = IoBlock> + '_ {
```
xiaguan, line 100: The implementation of this function seems a bit messy. I think we can refactor it. I feel like we don't need the abstraction of `IoBlock`; we can directly use Block and add other members inside it, along with corresponding initialization functions. If it's really necessary to use a data structure to represent the relationship between multiple blocks, I think `Slice` would be a better choice.

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -0,0 +1,812 @@
...
+    }
+
+    /// Load data from storage.
+    pub async fn load(
+        &self,
+        ino: INum,
+        offset: usize,
+        len: usize,
+        mtime: SystemTime,
+    ) -> Vec<IoBlock> {
```
xiaguan, line 173: Same above.

xiaguan (approved):

PR #443, opened 2023-12-13
docs: add CSI test doc

xiaguan:


Reto911 (approved):
The documentation and scripts are OK, and I tried to make the doc polished and fixed some grammar issues in it. You can sqaush the latest commit into your commits (or just leave it here), and update you branch to be based on master.

PR #444, opened 2023-12-25
reafctor(metadata) : use txn to do operations

xiaguan:
This pull request introduces **transactional mechanisms** to the metadata 's **read-modify-write** operations to ensure **atomicity**  and prevent race conditions. 

Over the coming days, I plan to refactor some of the longer functions.

### DIFF
--- a/src/async_fuse/memfs/s3_metadata.rs
+++ b/src/async_fuse/memfs/s3_metadata.rs
@@ -14,7 +14,7 @@ use nix::errno::Errno;
 use nix::sys::stat::SFlag;
 use parking_lot::RwLock as SyncRwLock; // conflict with tokio RwLock
 use tokio::sync::Mutex;
-use tracing::{debug, info, instrument, warn};
+use tracing::{debug, info, instrument};
 
 use super::cache::{GlobalCache, IoMemBlock};
 use super::dir::DirEntry;
@@ -22,9 +22,7 @@ use super::dist::client as dist_client;
 use super::dist::server::CacheServer;
 use super::fs_util::{self, FileAttr, NEED_CHECK_PERM};
 use super::id_alloc_used::INumAllocator;
-#[cfg(feature = "abi-7-23")]
-use super::kv_engine::MetaTxn;
-use super::kv_engine::{KVEngine, KVEngineType, KeyType, ValueType};
+use super::kv_engine::{KVEngine, KVEngineType, KeyType, MetaTxn, ValueType};
 use super::metadata::{error, MetaData, ReqContext};
 use super::node::Node;
 use super::s3_node::S3Node;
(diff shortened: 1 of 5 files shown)


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -258,32 +262,14 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
     #[instrument(skip(self), err, ret)]
     async fn open(&self, context: ReqContext, ino: u64, flags: u32) -> DatenLordResult<RawFd> {
-        let node = self
-            .get_node_from_kv_engine(ino)
-            .await?
-            .ok_or_else(|| build_inconsistent_fs!(ino))?;
-        let o_flags = fs_util::parse_oflag(flags);
-        node.open_pre_check(o_flags, context.user_id, context.group_id)?;
         // TODO: handle open flags
         // <https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html>
```
rogercloud, line 274: Do we still need these comments?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -293,57 +279,50 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
-                inode.get_name(),
-                previous_count,
-                current_count,
-            );
-            self.set_node_to_kv_engine(ino, inode).await?;
-        };
-        self.try_delete_node(ino).await?;
-
+        retry_txn!(TXN_RETRY_LIMIT, {
+            let mut txn = self.kv_engine.new_meta_txn().await;
```
rogercloud, line 312: This function call is always there, shall we put this statement into the retry_txn macro?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -293,57 +279,50 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+            let inode = self.get_inode_from_txn(txn.as_mut(), ino).await?;
+            inode.dec_lookup_count_by(nlookup);
+            let is_deleted = self.delete_check(&inode).await?;
+            if is_deleted {
+                txn.delete(&KeyType::INum2Node(ino));
+            } else {
+                txn.set(
+                    &KeyType::INum2Node(ino),
+                    &ValueType::Node(inode.into_serial_node()),
+                );
```
rogercloud, line 322: I've seen this block twice in this PR. Please put them into a function.

rogercloud (commented):
This PR removes some debug statements. I agree that these are not debug info but trace info. However, we should leave them there  on `trace` level or at least one for each function. They're useful when tracing is turned on.

xiaguan:
> This PR removes some debug statements. I agree that these are not debug info but trace info. However, we should leave them there on `trace` level or at least one for each function. They're useful when tracing is turned on.

The `tracing::instrument` will do the same thing. IIt will print the args first on'DEBUG' level . Then it will print the result  log on `INFO` level when  `result.is_ok()` , print error log on `ERROR` level when `result.is_err()`. 

Due to this resaon  , most of the preview logs are duplicated.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: Ok, I will refactor it into a new funcion.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: I will keep this statement. It can enhance readability. Meanwhile, the semantics of `retry_txn` is just to retry a transaction, and I think this way is better.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: We might use this later when we build the open-to-close cache.

rogercloud (approved):

PR #447, opened 2023-12-29
chore : use compact log format

xiaguan:
The previous logging format was too redundant. Use a more efficient format.


xiaguan:
Close.
A better pull request about logging is comming.

PR #453, opened 2024-01-03
refactor(log): Use compact log format

xiaguan:
In this pull request, I refactor the log datenlord print. Make it more compact, and more easier to process and read.

### DIFF
--- a/src/async_fuse/fuse/session.rs
+++ b/src/async_fuse/fuse/session.rs
@@ -13,7 +13,7 @@ use crossbeam_utils::atomic::AtomicCell;
 use nix::errno::Errno;
 use nix::sys::stat::SFlag;
 use nix::unistd;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, instrument};
 
 use super::context::ProtoVersion;
 use super::file_system::FileSystem;
@@ -446,12 +446,13 @@ impl<F: FileSystem + Send + Sync + 'static> Session<F> {
 /// This calls the appropriate filesystem operation method for the
 /// request and sends back the returned reply to the kernel
 #[allow(clippy::too_many_lines)]
+#[instrument(name="request",skip(req,fd, fs), fields(fuse_id =req.unique(),ino=req.nodeid()),ret)]
 async fn dispatch<'a>(
     req: &'a Request<'a>,
     fd: RawFd,
     fs: Arc<dyn FileSystem + Send + Sync + 'static>,
 ) -> nix::Result<usize> {
-    match *req.operation() {
+    let result = match *req.operation() {
         // Filesystem initialization
         Operation::Init { .. } => panic!("FUSE should have already initialized"),
 
@@ -808,7 +809,9 @@ async fn dispatch<'a>(
         Operation::CuseInit { arg } => {
             panic!("unsupported CuseInit arg={arg:?}");
         }
-    }
+    };
+
+    result
 }
 
 /// Replies ENOSYS
--- a/src/async_fuse/memfs/mod.rs
+++ b/src/async_fuse/memfs/mod.rs
@@ -143,7 +143,6 @@ const MAX_NAME_LEN: usize = 255;
 
 /// Check the name length is valid or not
 pub fn check_name_length(name: &str) -> DatenLordResult<()> {
-    debug!("check name length = {}", name.chars().count());
     if name.len() > MAX_NAME_LEN {
         error!("name = {} is too long ,size = {}", name, name.len());
         build_error_result_from_errno(Errno::ENAMETOOLONG, "name too long ".to_owned())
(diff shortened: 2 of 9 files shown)


src/async_fuse/memfs/mod.rs
```diff
@@ -238,22 +237,8 @@ impl<M: MetaData + Send + Sync + 'static> FileSystem for MemFs<M> {
         };
         let lookup_res = self.metadata.lookup_helper(context, parent, name).await;
         match lookup_res {
-            Ok((ttl, fuse_attr, generation)) => {
-                debug!(
-                    "lookup() successfully found the node name={:?} \
-                        under parent ino={}, the attr={:?}",
-                    name, parent, &fuse_attr,
-                );
```
Reto911, line 246: You removed these debug informations, is it because `tracing` has done it? If so, shall we double check those `debug!` in `S3Metadata`, and remove unneccessary ones?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -52,7 +52,7 @@ const MY_TTL_SEC: u64 = 3600; // TODO: should be a long value, say 1 hour
 const MY_GENERATION: u64 = 1; // TODO: find a proper way to set generation
 #[allow(dead_code)]
```
Reto911, line 53: By the way, we can remove this `dead_code` mark now, as the `TXN_RETRY_LIMIT` is commonly used in any ABI versions.

src/common/logger.rs
```diff
@@ -63,7 +63,7 @@ pub fn init_logger(role: LogRole) {
         .with_target("h2", Level::WARN)
         .with_target("tower", Level::WARN)
         .with_target("datenlord::async_fuse::fuse", Level::INFO)
-        .with_target("", Level::DEBUG);
+        .with_target("", Level::INFO);
```
Reto911, line 66: Would it be better to set `DEBUG` level in debug mode while `INFO` level in release mode?

Reto911 (commented):
Be care of the decreasing coverage, perhaps we should double check and remove the unnessessary `debug!` calls, and add `#[instrument]` for those functions that need to print tracing logs.

src/async_fuse/fuse/session.rs
```diff
@@ -446,12 +446,13 @@ impl<F: FileSystem + Send + Sync + 'static> Session<F> {
 /// This calls the appropriate filesystem operation method for the
 /// request and sends back the returned reply to the kernel
 #[allow(clippy::too_many_lines)]
+#[instrument(name="request",skip(req,fd, fs), fields(fuse_id =req.unique(),ino=req.nodeid()),ret)]
```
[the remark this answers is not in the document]
xiaguan: I don't think it's necessary because there's another layer of `#[instrument()]` on the metadata.

src/async_fuse/memfs/mod.rs
xiaguan: There are too many logs to check; I only dealt with the ones I could see.

src/common/logger.rs
xiaguan: I think debug mode should be enabled when a bug is discovered and needs debugging. By default, I prefer info because INFO logs are concise and efficient.

Reto911: So when do we enable DEBUG level? Only in debugging? And `debug!` should never exist in formal codes?

xiaguan: It's acceptable to have a certain amount of `!debug` logs, but only a small quantity. Compared to `!debug`, I'd prefer to see more `debug_assert!()` in the project. Currently, even running a simple load might generate 200,000 lines of info logs.

Reto911: > It's acceptable to have a certain amount of `!debug` logs, but only a small quantity. Compared to `!debug`, I'd prefer to see more `debug_assert!()` in the project. Currently, even running a simple load might generate 200,000 lines of info logs.

Make sense. Then we have more reasons to check the unneccessary `debug!` in metadata, remove them or replace then with `info!`. Maybe it's not needed to be done in this PR, but we should take cares of them.

src/async_fuse/memfs/mod.rs
Reto911: Please create an issue to track this problem.

xiaguan: #459

src/async_fuse/memfs/s3_node.rs
```diff
@@ -531,6 +538,13 @@ impl<S: S3BackEnd + Sync + Send + 'static> Node for S3Node<S> {
...
     /// Decrease node open count
     fn dec_open_count(&self) -> i64 {
+        if self.open_count.load(Ordering::Acquire) == 0 {
+            warn!(
+                "dec_open_count() found open_count is 0, ino={}, name={}",
+                self.get_ino(),
+                self.get_name()
+            );
+        }
         debug_assert!(self.open_count.load(Ordering::Acquire) > 0);
```
Reto911, line 548: There is a gap in the atomic operation between `warn` and `debug_assert`, is it OK?

Reto911 (approved):

src/async_fuse/memfs/s3_node.rs
xiaguan: It's OK. Lately ,the s3node is not thread safe since each s3node is owned by one operation.

PR #454, opened 2024-01-04
feat: Introduce write back for `MemoryCache`

Reto911:
In this PR, we introduced `MemoryCacheBuilder` to configure and build a `MemoryCache`, and write-back or write-through is a configurable feature for `MemoryCache` now. Besides, we introduced a background task to do evict and write back smoothly.


src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -2,27 +2,285 @@
...
+    storage: Weak<MemoryCache<P, S>>,
+    mut command_receiver: mpsc::Receiver<Command>,
+    command_queue_limit: usize,
+) where
+    P: EvictPolicy<BlockCoordinate> + Send + Sync,
+    S: Storage + Send + Sync,
+{
+    let period = interval;
+    let mut interval = tokio::time::interval(period);
+    let (a, b) = limit;
```
xiaguan, line 61: Rename ‘a' and 'b' for readbility.

Reto911: For example, `numerator` and `denominator`? I'm sorry that I can't give a good name to them. They are just two components of a fraction.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -2,27 +2,285 @@
...
+}
+
+/// A function to run a write back task.
+///
+/// This task will smoothly write blocks back to the backend
+/// when the cache runs in write-back policy.
+///
+/// And this task will evict blocks to the backend if the number of blocks hits the soft limit,
+/// whether the cache runs in write-back policy or not.
+async fn write_back_task<P, S>(
```
xiaguan, line 49: This function is too long. It should be broken down into three sub-functions:

1. The update of the period.
2. The handling of commands, each type of command should have its own handling function.
3. Is the handling of an expired interval also considered a type of command?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -2,27 +2,285 @@
...
+    limit: (usize, usize),
```
xiaguan, line 50: Replace `limit` with a simple structure; using a tuple is less readable.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: 3. No, it's not a type of command. A command is just a message sent from the cache. But put its handling into a sub-function is a good idea.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: But how can I name its fields🤔

xiaguan:
Let's consider adding a 100ms sleep to every operation in mock.rs (to simulate real delay, the bigger the better for testing. It can be optional). 

Eventually, we want to test the **latency** of the store_manager, for instance, how long store_blocks should take to return (write_back, write_through), and at the same time, read, truncate, etc. We need to have a quantitative understanding of its latency and throughput. What is the latency for each operation? How high can the throughput be? 

In this PR, we can just do a simple latency test.

src/async_fuse/memfs/cache/memory_cache.rs
xiaguan: Abstract it into a structure, and then provide an example in the comment documentation. `a`, `b`, and `(usize, usize)` have virtually no readability.

xiaguan:
The new tests added are quite good :+1: . However, they might have overlooked the path for updates (>= 200ms,read and write). 

Regarding the tests for write-back, we can sleep for about 100ms in the test, then call `Storage::flush()` and observe the latency.

src/async_fuse/memfs/cache/policy/lru.rs
```diff
@@ -53,6 +49,14 @@ impl<K: Clone + Hash + Eq> EvictPolicy<K> for LruPolicy<K> {
             true
         }
     }
+
+    fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    fn size(&self) -> usize {
+        self.inner.read().len()
```
xiaguan, line 58: I'm not quite sure if we need to expand our lock scope and switch to a read-write lock just for a size() interface. Given that the read intervals are very brief and writing is the more common load, I think sticking with a regular lock might be better.

xiaguan: At the same time, some comments need to be completed. Below are comments written with the help of GPT-4, which you can modify based on the above.
```rust
// Standard library imports for hashing and locking mechanisms.
use std::hash::Hash;
use hashlink::LinkedHashSet;
use parking_lot::RwLock;

// Import the EvictPolicy trait from the parent module.
use super::EvictPolicy;

/// The evict policy based on LRU (Least Recently Used).
/// It prioritizes removing items that were accessed the least recently.
#[derive(Debug)]
pub struct LruPolicy<K> {
    /// A thread-safe LinkedHashSet that stores the keys in insertion order.
    /// It is wrapped in a RwLock to allow safe concurrent access.
    inner: RwLock<LinkedHashSet<K>>,
    /// The maximum number of items the LRU policy can hold.
    capacity: usize,
}

impl<K: Hash + Eq> LruPolicy<K> {
    /// Create a new `LruPolicy` with the given `capacity`.
    /// This initializes the inner LinkedHashSet with the specified capacity.
    #[must_use]
    pub fn new(capacity: usize) -> Self {
        LruPolicy {
            inner: RwLock::new(LinkedHashSet::with_capacity(capacity)),
            capacity,
        }
    }
}

impl<K: Clone + Hash + Eq> EvictPolicy<K> for LruPolicy<K> {
    /// Update the position of the given key to mark it as recently used.
    /// If the key exists, it is moved to the back of the set, indicating recent use.
    fn touch(&self, key: &K) {
        self.inner.write().to_back(key);
    }

    /// Remove and return the least recently used item.
    /// This is typically the first item in the set.
    fn evict(&self) -> Option<K> {
        let mut lru = self.inner.write();

        lru.pop_front()
    }

    /// Attempt to insert the key into the set.
    /// If the set is at capacity and does not contain the key, return false.
    /// If the key is inserted successfully or already exists, return true.
    fn try_put(&self, key: K) -> bool {
        let mut lru = self.inner.write();
        let len = lru.len();

        if !lru.contains(&key) && len == self.capacity {
            false
        } else {
            lru.insert(key);
            true
        }
    }

    /// Return the capacity of the LRU policy.
    fn capacity(&self) -> usize {
        self.capacity
    }

    /// Return the current number of items in the LRU policy.
    fn size(&self) -> usize {
        self.inner.read().len()
    }
}
```

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -2,27 +2,373 @@
...
-type FileCache = Arc<RwLock<StdHashMap<BlockId, Block>>>;
-
 /// Merge the content from `src` to `dst`. This will set `dst` to be dirty.
 fn merge_two_blocks(src: &Block, dst: &mut Block) {
-    dst.set_dirty();
+    dst.set_dirty(true);
     dst.update(src);
 }
 
+/// A command sent from `MemoryCache` to the write back task.
```
xiaguan, line 25: Split to another rust file with write_back_task related.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -2,27 +2,373 @@
...
+
+    use super::{Block, BlockId, EvictPolicy, INum, MemoryCache, SoftLimit, StdHashMap, Storage};
+    use crate::async_fuse::memfs::cache::BlockCoordinate;
+    /// Calculate a new period for write back task's interval.
+    pub(super) fn calculate_period(
+        pending_length: usize,
+        command_queue_limit: usize,
+        basic_period: Duration,
+    ) -> Duration {
+        if pending_length > command_queue_limit.overflow_div(4).overflow_mul(3) {
```
xiaguan, line 59: Make these msgic numbers to constants.

Reto911: They are also fractions. I will write more comments to make it clearer, but I think there is no need to make them to constants.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: It's not convenient to access private fields in `MemoryCache` from write back task, and make those fields `pub` is worse practice, therefore, I insist to hold them in the same file.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -3,26 +3,139 @@
...
-
 /// Merge the content from `src` to `dst`. This will set `dst` to be dirty.
 fn merge_two_blocks(src: &Block, dst: &mut Block) {
-    dst.set_dirty();
+    dst.set_dirty(true);
     dst.update(src);
 }
 
+/// A builder to configure and build a `MemoryCache`.
+pub struct MemoryCacheBuilder<P, S> {
```
xiaguan, line 24: This should be moved to another file, as each file should maintain its own minimal responsibility.

src/async_fuse/memfs/cache/mock.rs
```diff
@@ -18,18 +21,28 @@ pub struct MemoryStorage {
     flushed: Mutex<HashSet<INum>>,
     /// The size of block
     block_size: usize,
+    /// The latency of all operations
+    latency: Duration,
 }
 
 impl MemoryStorage {
     /// Creates a memory storage with block size.
```
xiaguan, line 29: We should `#[cfg(test)]` for the mock struct ?

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -3,26 +3,139 @@
...
+            NonZeroUsize::new(5).unwrap_or_else(|| unreachable!("5 is not zero.")),
+        );
+        Self {
+            policy,
+            backend,
+            block_size,
+            write_through: true,
+            limit,
+            interval: Duration::from_millis(100),
+            command_queue_limit: 1000,
```
xiaguan, line 59: Represent all static numbers using constants。

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: `MemoryCacheBuilder` is highly related with `MemoryCache`, so I insist on leaving it in the same file.

src/async_fuse/memfs/cache/memory_cache.rs
```diff
@@ -238,18 +398,108 @@ where
     }
 
     async fn flush(&self, ino: INum) {
+        if !self.write_through {
+            {
+                let mut pending_write_back = self.pending_write_back.write().await;
```
xiaguan, line 403: So, do we need to hold the write lock during the IO time flushed out in a certain file?

src/async_fuse/memfs/cache/memory_cache.rs
xiaguan: Then, let's put them in a subfolder named `memory_storage` or something, with files like `mod.rs`, `memory_cache.rs`, `build.rs`, and `test.rs`. This file is already 1000 lines long.

xiaguan: Tests for other modules might also need to be split. For example, `storage_manager` has 700 lines of tests and 300 lines of source code.

src/async_fuse/memfs/cache/write_back_task.rs
```diff
@@ -0,0 +1,239 @@
...
+                    Some(Command::Store { coord: BlockCoordinate(ino, block_id), block, sender }) => {
+                        // Add the block to `lru_queue` and `pending_blocks`
+                        lru_queue.insert(BlockCoordinate(ino, block_id));
+                        pending_blocks.entry(ino).or_default().insert(block_id, (block, sender));
+                    }
+                    Some(Command::Flush(ino)) => {
+                        if let Some(cache) = storage.upgrade() {
+                            flush_file(&mut lru_queue, &mut pending_blocks, cache, ino).await;
+                        } else {
+                            return;
```
xiaguan, line 197: It should panic here.

src/async_fuse/memfs/cache/write_back_task.rs
```diff
@@ -0,0 +1,239 @@
...
+                        }
+                    },
+                    Some(Command::FlushAll(sender)) => {
+                        if let Some(cache) = storage.upgrade() {
+                            flush_all(&mut lru_queue, &mut pending_blocks, cache, sender).await;
+                        } else {
+                            return;
```
xiaguan, line 204: Same above.

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -768,4 +802,139 @@ mod tests {
...
+        #[tokio::test]
+        async fn test_write_back_flush_latency() {
+            let storage = create_storage_with_latency(false);
+            let mut mtime = SystemTime::now();
+            for block_id in 0..8 {
+                let offset = block_id.overflow_mul(BLOCK_SIZE_IN_BYTES);
+                mtime = storage.store(0, offset, BLOCK_CONTENT, mtime).await;
+            }
+
+            tokio::time::sleep(Duration::from_millis(150)).await;
```
xiaguan, line 932: 150 -> 900
At the end of assert ,asssert!(lantency < 10ms).

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: No, we don't. The write lock here is to ensure that no more blocks can be add to the `pending_write_back` queue, before we send a `Flush` command to the task. As we have sent the command, a file-level queue will be removed from `pending_write_back`, then the write lock can be released.

src/async_fuse/memfs/cache/memory_cache.rs
Reto911: > Then, let's put them in a subfolder named `memory_storage` or something, with files like `mod.rs`, `memory_cache.rs`, `build.rs`, and `test.rs`. This file is already 1000 lines long.

This is a good idea， I will handle it later. However, I think it's not good to handle `storage_manager` related issues in this PR.

src/async_fuse/memfs/cache/write_back_task.rs
Reto911: A `Flush` command was sent, but the `MemoryCache` is dropped before the task can handle the command, then the upgrade will fail. Do you think this should be an error? Or this is unreachable?

xiaguan (dismissed):

xiaguan (approved):

PR #457, opened 2024-01-09
feat: Introduce backend and add configurable items

Reto911:
In this PR, we introduced the backend with `openDAL`, and added new items to configure the storage module.

The patches of config items are listed below:
1. Renamed one of the variant of storage type from `None` to `Fs`, and changed its member from a `S3StorageConfig` to a String representing the root of filesystem based backend.
2. Added `--storage-mem-cache-capacity`, `--storage-mem-cache-command-limit`, `--storage-mem-write-back` and `--storage-mem-cache-soft-limit` as the cocnfig of memory cache.
3. Removed `--storage-cache-capacity`, it's for the legacy `GlobalCache`, which will be removed soon.

### DIFF
--- a/scripts/ci/csi-e2e-test.sh
+++ b/scripts/ci/csi-e2e-test.sh
@@ -40,9 +40,3 @@ kubernetes/test/bin/ginkgo -v -failFast -failOnPending -debug -focus='External.S
 # Run [Disruptive] test in serial and separately
 kubernetes/test/bin/ginkgo -v -failFast -failOnPending -debug -focus='External.Storage.*(\[Feature:|\[Disruptive\]|\[Serial\])' kubernetes/test/bin/e2e.test -- -v=5 -kubectl-path=`which kubectl` -kubeconfig=`realpath $K8S_CONFIG` -storage.testdriver=`realpath $E2E_TEST_CONFIG`
 /bin/sh /tmp/clean_up_mount_dir.sh $CONFIG_KIND $NODE_APP_LABEL $DATENLORD_NAMESPACE false
-
-# # Quick Start Test
-sed -e 's/e2e_test/latest/g' $CONFIG_KIND > $CONFIG_DOCKERHUB
-kubectl apply -f $CONFIG_DOCKERHUB
-kubectl wait --for=condition=Ready pod -l app=$CONTROLLER_APP_LABEL -n $DATENLORD_NAMESPACE --timeout=120s
-kubectl wait --for=condition=Ready pod -l app=$NODE_APP_LABEL -n $DATENLORD_NAMESPACE --timeout=120s
--- a/src/async_fuse/memfs/cache/backend.rs
+++ b/src/async_fuse/memfs/cache/backend.rs
@@ -2,7 +2,9 @@
 
 use async_trait::async_trait;
 use clippy_utilities::OverflowArithmetic;
+use datenlord::config::{StorageParams, StorageS3Config};
 use futures::{stream, AsyncReadExt, AsyncWriteExt, StreamExt};
+use opendal::services::{Fs, S3};
 use opendal::{ErrorKind, Operator};
 
 use super::{Block, Storage};
@@ -16,6 +18,58 @@ fn get_file_path(ino: INum) -> String {
 /// Get block path by `ino` and `block_id`
 fn get_block_path(ino: INum, block_id: usize) -> String {
     format!("{ino}/{block_id}.block")
+}
+
+/// A builder to build `BackendWrapper`.
+#[derive(Debug)]
+pub struct BackendBuilder {
+    /// The storage config
+    config: StorageParams,
+    /// The size of a block
+    block_size: usize,
+}
+
+impl BackendBuilder {
+    /// Create a backend builder.
+    #[must_use]
+    pub fn new(config: StorageParams, block_size: usize) -> Self {
+        Self { config, block_size }
+    }
+
+    /// Build the backend.
+    pub fn build(self) -> opendal::Result<Backend> {
+        let BackendBuilder { config, block_size } = self;
+
+        let operator = match config {
+            StorageParams::S3(StorageS3Config {
+                ref endpoint_url,
+                ref access_key_id,
+                ref secret_access_key,
+                ref bucket_name,
+            }) => {
+                let mut builder = S3::default();
+
+                builder
+                    .endpoint(endpoint_url)
+                    .access_key_id(access_key_id)
+                    .secret_access_key(secret_access_key)
+                    .region("auto")
+                    .bucket(bucket_name);
+
+                Operator::new(builder)?.finish()
+            }
+            StorageParams::Fs(ref root) => {
+                let mut builder = Fs::default();
+                builder.root(root);
+                Operator::new(builder)?.finish()
+            }
+        };
+
+        Ok(Backend {
+            operator,
+            block_size,
+        })
+    }
 }
 
 /// The backend wrapper of `openDAL` operator.
--- a/src/async_fuse/memfs/cache/backend.rs
+++ b/src/async_fuse/memfs/cache/backend.rs
@@ -1,0 +1,429 @@
+//! The implementation of backend.
+
+use async_trait::async_trait;
+use clippy_utilities::OverflowArithmetic;
+use futures::{stream, AsyncReadExt, AsyncWriteExt, StreamExt};
+use opendal::{ErrorKind, Operator};
+
+use super::{Block, Storage};
+use crate::async_fuse::fuse::protocol::INum;
+
+/// Get file path by `ino`
+fn get_file_path(ino: INum) -> String {
+    format!("{ino}/")
+}
+
+/// Get block path by `ino` and `block_id`
+fn get_block_path(ino: INum, block_id: usize) -> String {
+    format!("{ino}/{block_id}.block")
+}
+
+/// The backend wrapper of `openDAL` operator.
+#[derive(Debug)]
+pub struct Backend {
+    /// The inner `Operator`
+    operator: Operator,
+    /// Block size
+    block_size: usize,
+}
+
+impl Backend {
+    /// Create a new backend
+    #[must_use]
+    pub fn new(operator: Operator, block_size: usize) -> Self {
+        Self {
+            operator,
+            block_size,
+        }
+    }
+}
+
+#[async_trait]
+impl Storage for Backend {
+    async fn load_from_self(&self, ino: INum, block_id: usize) -> Option<Block> {
+        let mut block = Block::new_zeroed(self.block_size);
+
+        if let Err(e) = self
+            .operator
+            .reader(&get_block_path(ino, block_id))
+            .await
+            .unwrap_or_else(|e| {
+                panic!("Failed to get a reader where ino={ino} and block={block_id}: {e}")
+            })
+            .read(block.make_mut_slice())
+            .await
+        {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                None
+            } else {
+                panic!(
+                    "Failed to load a block from backend where ino={ino} and block={block_id}: {e}"
+                );
+            }
+        } else {
+            Some(block)
+        }
+    }
+
+    async fn load_from_backend(&self, _: INum, _: usize) -> Option<Block> {
+        // This storage has no backend.
+        None
+    }
+
+    async fn cache_block_from_backend(&self, _: INum, _: usize, _: Block) {
+        unreachable!("This storage has no backend, and has no cache.");
+    }
+
+    async fn store(&self, ino: INum, block_id: usize, block: Block) {
+        let path = get_block_path(ino, block_id);
+
+        let block_start = block.start();
+        let block_end = block.end();
+
+        if block_start == 0 && block_end == self.block_size {
+            // To store a whole block
+            let mut writer = self.operator.writer(&path).await.unwrap_or_else(|e| {
+                panic!(
+                    "Failed to get a writer to backend where ino={ino} and block={block_id}: {e}"
+                )
+            });
+            writer.write_all(block.as_slice()).await.unwrap_or_else(|e| panic!("Failed to store a block to backend where ino={ino} and block={block_id}: {e}"));
+            writer.close().await.unwrap_or_else(|e| {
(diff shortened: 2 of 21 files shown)


scripts/ci/csi-e2e-test.sh
```diff
@@ -42,7 +42,7 @@ kubernetes/test/bin/ginkgo -v -failFast -failOnPending -debug -focus='External.S
 /bin/sh /tmp/clean_up_mount_dir.sh $CONFIG_KIND $NODE_APP_LABEL $DATENLORD_NAMESPACE false
 
 # # Quick Start Test
```
rogercloud, line 44: Just remove them if we really don't need them.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,493 @@
...
+                    "Failed to load a block from backend where ino={ino} and block={block_id}: {e}"
+                );
+            }
+        } else {
+            Some(block)
+        }
+    }
+
+    async fn load_from_backend(&self, _: INum, _: usize) -> Option<Block> {
+        // This storage has no backend.
```
rogercloud, line 131: Why this function returns None, while the `cache_block_from_backend` just panics? Can we make them consistent?

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,493 @@
...
+
+    async fn remove(&self, ino: INum) {
+        self.operator
+            .remove_all(&Self::get_file_path(ino))
+            .await
+            .unwrap_or_else(|e| panic!("Failed to remove a file of ino={ino}: {e}"));
+    }
+
+    async fn invalidate(&self, _: INum) {
+        // This storage has no cache, therefore, its contents cannot be
```
rogercloud, line 190: The same question for these three functions as above.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,493 @@
...
+        let paths =
+            stream::iter(to_block..from_block).map(|block_id| Self::get_block_path(ino, block_id));
+
+        let file_path = Self::get_file_path(ino);
+
+        let file_exists = self
+            .operator
+            .is_exist(&file_path)
+            .await
+            .unwrap_or_else(|e| panic!("Failed to check if file ino={ino} exists: {e}"));
```
rogercloud, line 214: Can we treat this error as "not exist" instead of panic?

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,493 @@
...
+        for block_id in 1..4 {
+            let loaded = backend.load(0, block_id).await;
+            assert!(loaded.is_none());
+        }
+
+        backend.truncate(0, 1, 0, BLOCK_SIZE_IN_BYTES).await;
+
+        let file_path = backend_root.join("0");
+        let file_dir_exist = fs::try_exists(&file_path).await.unwrap();
+        assert!(!file_dir_exist);
```
rogercloud, line 410: What's purpose of this block?

src/async_fuse/memfs/mod.rs
```diff
@@ -1,5 +1,5 @@
 //! The implementation of user space file system
-mod cache;
+pub mod cache;
```
rogercloud, line 2: Where is the mod used outside of the memfs module?

src/async_fuse/mod.rs
```diff
@@ -42,7 +45,7 @@ pub async fn start_async_fuse(
         StorageParams::S3(_) => {
             let fs: memfs::MemFs<memfs::S3MetaData<S3BackEndImpl>> = memfs::MemFs::new(
                 &args.mount_dir,
-                args.storage_config.cache_capacity,
+                CACHE_DEFAULT_CAPACITY,
```
rogercloud, line 48: Do we still have a way to set the cache capacity? If yes, why do we pass the default capacity in here?

Reto911: This capacity is used by `GlobalCache`, which will be removed soon. And I removed the related config, so there are not direct ways to set it. If we still need to set the capacity for it, I can try to use the new capacity config (in blocks) defined for the new `MemoryCache`, to handle this temporarily.

src/async_fuse/memfs/cache/backend.rs
Reto911: `load_from_backend` will always be called when `load_from_self` returns `None`; and `cache_block_from_backend` will never be called if `load_from_backend` always returns `None`. This is decided by the default implementation of `load()` method.

I can remove the `unreachable!` in `cache_block_from_backend`, but I think panic in a function that will never be called is more semantically clear.

src/async_fuse/memfs/cache/backend.rs
Reto911: Each storage layer will also calls `invalidate()` method of its backend (see the implementation of `MemoryCache::invalidate()`), therefore, this method is always called too, so I can't raise a panic here.

src/async_fuse/memfs/cache/backend.rs
Reto911: Ok, I will fix it later.

src/async_fuse/memfs/cache/backend.rs
Reto911: If a file is truncated to `block_id = 0`, the whole file will be removed from the storage. If so, the directory representing this file will be removed directly. And this block is to test if the whole dir is removed properly.

I'm sorry for the unclear codes, I will write more comments for the test cases.

src/async_fuse/memfs/mod.rs
Reto911: No usages outside of the `memfs` module for now. But when we refactor the `S3Metadata` with it, the mod `async_fuse` and `async_fuse::test` will need to use this mod, so I need to make it public.

Finally, this mod will be moved to the root of this crate (parallel to mod `async_fuse`), and renamed to `storage`.

src/async_fuse/memfs/cache/backend.rs
xiaguan: I think a panic is better here to prevent accidental operations.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+                    .endpoint(endpoint_url)
+                    .access_key_id(access_key_id)
+                    .secret_access_key(secret_access_key)
+                    .region("auto")
+                    .bucket(bucket_name);
+
+                Operator::new(builder)?.finish()
+            }
+            StorageParams::Fs(ref root) => {
+                let root = if root.is_empty() {
```
xiaguan, line 53: It’s better to make this value the default in the configuration system.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+impl BackendBuilder {
+    /// Create a backend builder.
+    #[must_use]
+    pub fn new(config: StorageParams, block_size: usize) -> Self {
+        Self { config, block_size }
+    }
+
+    /// Build the backend.
+    pub fn build(self) -> opendal::Result<BackendWrapper> {
+        let BackendBuilder { config, block_size } = self;
```
xiaguan, line 32: Is there a valid range for `block_size`? eg : [4k,4mb]  If so, use `debug_assert!` to enforce it

src/async_fuse/memfs/cache/backend.rs
Reto911: > I think a panic is better here to prevent accidental operations.

But I can't. This method will be called by the default implementation of `load()`, when `load_from_self` returns `None`. That means, a calling to this method is a normal situation, while the calling to `cache_block_from_backend` is not. So I can't raise a panic in this method.

src/async_fuse/memfs/cache/backend.rs
Reto911: Ok, I will do it later.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+        Ok(BackendWrapper {
+            operator,
+            block_size,
+        })
+    }
+}
+
+/// The backend wrapper of `openDAL` operator.
+#[derive(Debug)]
+pub struct BackendWrapper {
```
xiaguan, line 75: I prefer the name `Backend` because we aren’t wrapping anything.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+    #[must_use]
+    pub fn new(operator: Operator, block_size: usize) -> Self {
+        Self {
+            operator,
+            block_size,
+        }
+    }
+
+    /// Get file path by `ino`
+    fn get_file_path(ino: INum) -> String {
```
xiaguan, line 93: These two functions actually have nothing to do with the backend; making them regular functions would suffice.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+    async fn load_from_backend(&self, _: INum, _: usize) -> Option<Block> {
+        // This storage has no backend.
+        None
+    }
+
+    async fn cache_block_from_backend(&self, _: INum, _: usize, _: Block) {
+        unreachable!("This storage has no backend, and has no cache.");
+    }
+
+    async fn store(&self, ino: INum, block_id: usize, block: Block) {
```
xiaguan, line 139: The `store` should return a result, as it cannot guarantee a 100% success rate in storing.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    /// Get block path by `ino` and `block_id`
+    fn get_block_path(ino: INum, block_id: usize) -> String {
+        format!("{ino}/{block_id}.block")
+    }
+}
+
+#[async_trait]
+impl Storage for BackendWrapper {
```
xiaguan, line 104: It seems that there are many code style issues in implementing this trait. Would it be better to implement a different interface?

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,491 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::indexing_slicing)]
+mod tests {
```
xiaguan, line 244: Add some pessimistic tests that should either error out or panic."

src/config/config.rs
```diff
@@ -40,19 +40,49 @@ pub struct Config {
...
+/// Memory cache config
+#[derive(Debug, Parser)]
+pub struct MemoryCacheConfig {
+    /// The capacity of memory cache in blocks, default is 16384 blocks.
+    #[clap(
+        long = "storage-mem-cache-capacity",
+        value_name = "VALUE",
+        default_value_t = 16384
+    )]
+    pub capacity: usize,
```
xiaguan, line 66: Use the size of memory instead of the number of stored blocks.

src/config/config.rs
```diff
@@ -40,19 +40,49 @@ pub struct Config {
...
+    /// A flag whether the cache runs in write-back policy, default is false
+    #[clap(long = "storage-mem-cache-write-back")]
+    pub write_back: bool,
+    /// A soft limit of evict policy for write back task.
+    ///
+    /// It's a fraction with the form of `a,b`, which means that
+    /// the soft limit is `a/b` of the capacity.
+    ///
+    /// **Note**: `b` cannot be set to 0.
+    #[clap(long = "storage-mem-cache-soft-limit", default_value = "3,5")]
```
xiaguan, line 84: Using floating-point numbers here would provide a better experience for the user.

Reto911: I tried, but it's not easy. I can't cast a `usize` to `f64` safely, and of course it's not safe to cast `f64` to `usize`. If they are not in the same type, I can't multiply them together. Therefore I chose to use fraction to represent the soft limit.

src/config/config.rs
Reto911: Should we also provide an item to config the `block_size`? Or just leave it as a constant for now?

xiaguan: provide with default value.

src/async_fuse/memfs/cache/backend.rs
Reto911: Maybe, but I think it's not good to reimplement it now. Perhaps we can remove the default implementation, and abstract the `load` operation in another way. But anyway, it's not good to do it now, at least not good in this PR.

src/async_fuse/memfs/cache/backend.rs
Reto911: This will be done in the future, when we introduce error handling for `Storage`.

src/async_fuse/memfs/cache/backend.rs
Reto911: No, there is no a valid range. The only limitation is that all storage layers must be with the same `block_size`, which will be guarnteed on creating the `StorageManager`. And in the future, we will record `block_size` in metadata, and this field here will be removed.

src/async_fuse/memfs/cache/backend.rs
```diff
@@ -0,0 +1,611 @@
...
+                    .access_key_id(access_key_id)
+                    .secret_access_key(secret_access_key)
+                    .region("auto")
+                    .bucket(bucket_name);
+
+                Operator::new(builder)?.finish()
+            }
+            StorageParams::Fs(ref root) => {
+                let root = if root.is_empty() {
+                    warn!("Argument `--storage-fs-root` is not set while `--storage-type` is set to `fs`. using `/tmp/datenlord_backend` as default.");
```
xiaguan, line 64: it's unreachable?

Reto911: Sorry, I forgot to handle this.

Reto911:
> The rest of the code looks good to me ，LGTM.
> 
> Here are some discussions: I think the storage should provide simpler operation **primitives**, similar to a block device, which is not aware of the details of the upper-layer file system.

I agree with it.

xiaguan (approved):

PR #461, opened 2024-01-11
fix: make write back concurrently

Reto911:
In this PR, we made flush and store operations concurrent.

### DIFF
--- a/src/async_fuse/memfs/cache/memory_cache/cache.rs
+++ b/src/async_fuse/memfs/cache/memory_cache/cache.rs
@@ -1,6 +1,8 @@
 //! The `MemoryCache` implementation.
 
 use std::collections::{BTreeSet, HashMap as StdHashMap};
+use std::mem;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use async_trait::async_trait;
@@ -38,10 +40,10 @@ pub struct MemoryCache<P, S> {
     /// The block size
     block_size: usize,
     /// A flag that if the `MemoryCache` runs in writing-through policy
-    write_through: bool,
+    write_through: AtomicBool,
     /// The pending written-back blocks.
     /// TODO: receive a `Result` to handle the error
-    pending_write_back: RwLock<StdHashMap<INum, Vec<oneshot::Receiver<()>>>>,
+    pending_write_back: RwLock<StdHashMap<INum, StdHashMap<BlockId, oneshot::Receiver<()>>>>,
     /// A command sender for write back task
     command_sender: mpsc::Sender<Command>,
     /// A set of blocks to be removed, when a file is truncated.
@@ -65,10 +67,43 @@ impl<P, S> MemoryCache<P, S> {
             policy,
             backend,
             block_size,
-            write_through,
+            write_through: AtomicBool::new(write_through),
             pending_write_back: RwLock::default(),
             command_sender,
             truncate_records: HashMap::new(),
+        }
+    }
+
+    /// Check if the cache is running on wtire through policy.
+    fn write_through(&self) -> bool {
+        self.write_through.load(Ordering::Acquire)
+    }
+
+    /// Fallback to write through.
+    async fn fallback(&self)
+    where
+        S: Storage,
+    {
+        self.write_through.store(true, Ordering::Release);
+
+        // Ensure that all pending blocks are flushed to the backend.
+        let pending = mem::take(&mut *self.pending_write_back.write().await);
+        for (ino, pending_blocks) in pending {
+            if let Some(lock) = self.get_file_cache(ino) {
+                let cache = lock.read_owned().await;
+
+                for (block_id, receiver) in pending_blocks {
+                    if receiver.await.is_err() {
+                        // This block is not flushed to the backend successfully.
+                        // Flush it now.
+                        if let Some(block) = cache.get(&block_id) {
+                            self.backend.store(ino, block_id, block.clone()).await;
+                        }
+                        // If the block is not shown in the cache, it should
+                        // have been evicted to the backend.
+                    }
+                }
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/async_fuse/memfs/cache/memory_cache/cache.rs
```diff
@@ -306,6 +306,18 @@ where
...
     async fn remove(&self, ino: INum) {
+        if !self.write_through {
+            let mut pending_write_back = self.pending_write_back.write().await;
+            if self
+                .command_sender
+                .send(Command::Cancel(ino))
+                .await
+                .is_err()
+            {
+                warn!("Write back task is closed unexpectedly.");
```
rogercloud, line 317: Is warning enough, shall we fallback to the write through policy here?

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
```diff
@@ -54,24 +58,50 @@ fn calculate_period(
...
+    block_id: BlockId,
+    block: Block,
+    sender: oneshot::Sender<()>,
+) where
+    P: EvictPolicy<BlockCoordinate> + Send + Sync + 'static,
+    S: Storage + Send + Sync + 'static,
+{
+    cache.backend().store(ino, block_id, block).await;
+    sender.send(()).unwrap_or_else(|()| {
+        warn!("The receiver of pending block is closed unexpectedly.");
```
rogercloud, line 74: The same as above.

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -928,11 +928,10 @@ mod tests {
...
             }
 
-            // Wait for all blocks being flushed
-            tokio::time::sleep(Duration::from_millis(900)).await;
-
             let (latency, ()) = elapsed!(storage.flush(0).await);
-            assert!(latency.as_millis() < 10, "latency = {latency:?}");
+            // All blocks are flushed concurrently, therefore the latency of flushing
+            // should be a single operation latency (100ms) + some extra costs (< 10ms)
+            assert!(latency.as_millis() < 110, "latency = {latency:?}");
```
rogercloud, line 934: hard code the latency is not good. Is there any constant definition?

src/async_fuse/memfs/cache/memory_cache/cache.rs
Reto911: It's enough. This is to cancel the flushing of the file, if the write back task is closed, these blocks will not be flushed to the backend anymore.

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
Reto911: It's also enough. The sender here is to notify the caller (usually the FUSE operation handler) that this block is flushed to the backend. Although the Receiver held by the caller is dropped, the block has been flushed.

src/async_fuse/memfs/cache/storage_manager.rs
Reto911: Yes, it's not good. I will fix it later.

src/async_fuse/memfs/cache/memory_cache/cache.rs
Reto911: But I will double check the related codes in `MemoryCache`, to ensure that the block to be flushed will not be ignored when the write back task is closed unexpectedly.

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
Reto911: This is enough. The receivers here are held by `MemoryCache`, if the receivers are closed unexpectedly, the `MemoryCache` must be dropped, and the async FUSE task and operation handlers must quit. It's not bad, because anyway, the blocks in write back task have been flushed to the backend normally.

src/async_fuse/memfs/cache/memory_cache/cache.rs
```diff
@@ -65,13 +67,46 @@ impl<P, S> MemoryCache<P, S> {
...
+
+    /// Fallback to write through.
+    async fn fallback(&self)
+    where
+        S: Storage,
+    {
+        self.write_through.store(true, Ordering::Release);
+
+        // Ensure that all pending blocks are flushed to the backend.
+        let pending = mem::take(&mut *self.pending_write_back.write().await);
```
rogercloud, line 90: `mem::take` actually copy the whole data struture out to the stack. Is that what you really want?

rogercloud: I guess what you want is "release the lock as soon as possible".

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
```diff
@@ -32,205 +33,294 @@ pub enum Command {
...
+                    block,
+                    sender,
+                ));
+                self.task_queue.push_back(handle);
+            }
+        }
+    }
+
+    /// Flush a file, write blocks of it to backend immediately.
+    async fn flush_file(&mut self, cache: Arc<MemoryCache<P, S>>, ino: INum)
```
rogercloud, line 152: This function is quite similar with the later one, can we refactor them?

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
```diff
@@ -32,205 +33,294 @@ pub enum Command {
...
+    {
+        // `a` is a numerator while `b` is a denominator, which are to represent the
+        // soft limit. If `size > capacity * a / b`, the soft limit is considered to
+        // be hit. We use the equivalent inequation `size * b / a > capacity` to
+        // avoid overflowing.
+        let SoftLimit(a, b) = self.limit;
+        let b = b.get();
+
+        loop {
+            let capacity = cache.policy().capacity();
```
rogercloud, line 233: get a policy reference before the loop?

src/async_fuse/memfs/cache/memory_cache/cache.rs
Reto911: > I guess what you want is "release the lock as soon as possible".

Yes. And `mem::take` should do shallow copy (in move semantics), so I think that to `take` a `HashMap` is OK.

src/async_fuse/memfs/cache/memory_cache/write_back_task.rs
Reto911: Ther are similar now, but when we introduce `TaskManager` here, they will perform different behaviours: `flush_file` spawns tasks to flush blocks, and submits them into the `TaskManager`; while `flush_all` spawns those tasks and await them immediately, then send a signal to the caller in `MemoryCache` via `sender`. That's because `flush_all` is usually called once when shutdown, it's not able to spawn task with `TaskManager`.

xiaguan:
I won't approve this pull request. The crux of the issue lies in the potential errors in flush tasks, and simply degrading to a write-through strategy is not a solution. The fundamental problem is how to **manage flush tasks** and maintain per file I/O status.

xiaguan:
For example, if the error is caused by network issues, degrading to write-through won't solve the problem either.

Reto911:
> I won't approve this pull request. The crux of the issue lies in the potential errors in flush tasks, and simply degrading to a write-through strategy is not a solution. The fundamental problem is how to **manage flush tasks** and maintain per file I/O status.

Yes, this needs further consideration. So I think the concurrency stuffs should be done after we introduce some mechanism to handle the error and tasks (such as `Error` in `Storage` and `TaskManager` and so on). I think this PR should be suspended, before the related issues are resolved.

Reto911:
> For example, if the error is caused by network issues, degrading to write-through won't solve the problem either.

Ideally, the write back task should never exits, unless the whole system shutdown. So when we can make such a guarantee, all branch of the write back task exiting unexpectedly (for example, the `command_tx` closed early) should `panic!` or even be `unreachable!`. Any IO issues should not make the task panic or exit, while all errors should be returned to the `StorageManager` and `FileSystem` (or `S3Metadata` now), via the oneshot sender passed to the block flushing tasks.

Reto911:
Resolved in #493

PR #463, opened 2024-01-16
refactor(fs): Introduce storage for fs

Reto911:
In this PR, we refactored the `S3Metadata` with new `StorageManager`, thus the old `S3Backend`, distributed cache and `GlobalCache` are deprecated.

Because of the patch size, they will not be removed in this PR, causing the decreasing coverage.

### DIFF
--- a/src/async_fuse/memfs/s3_metadata.rs
+++ b/src/async_fuse/memfs/s3_metadata.rs
@@ -816,7 +816,7 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
                     } else {
                         let mut old_raw_node = old_parent_node.lock().await;
                         let mut new_raw_node = new_parent_node.lock().await;
-                        old_raw_node.get_dir_data_mut().remove(old_name);
+                        old_raw_node.remove_entry_for_rename(old_name);
                         new_raw_node.insert_entry_for_rename(DirEntry::new(
                             new_name.into(),
                             Arc::clone(old_entry.file_attr_arc_ref()),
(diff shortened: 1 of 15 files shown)


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -56,11 +60,13 @@ const TXN_RETRY_LIMIT: u32 = 10;
...
 #[derive(Debug)]
 #[allow(dead_code)]
-pub struct S3MetaData<S: S3BackEnd + Send + Sync + 'static> {
+pub struct S3MetaData<S: S3BackEnd + Send + Sync + 'static, St: Storage + Send + Sync + 'static> {
     /// S3 backend
     pub(crate) s3_backend: Arc<S>,
     /// Global data cache
     pub(crate) data_cache: Arc<GlobalCache>,
+    /// Storage manager
+    pub(crate) storage: Arc<StorageManager<St>>,
```
xiaguan, line 69: I don't think generics are necessary here. Generics should be more about having code run with different types at runtime, like in the case of `max` and `min` func . But in reality, after our configuration, `MetaData` is **no longer generic**. I believe it would be more appropriate to use associated types, or even to hard code the implementation.

In the final version, both Metadata and Node should be non-generic.

Reto911: Perhaps we can refactor all of them in the future, but generic (`St` here) is neccessary for `StorageManager`. If we introduce more storage layers in the future, we may have to define a very complex type named `MemoryCache<LruPolicy<BlockCoordinate>, DiskCache<FifoPolicy<BlockCoordinate>, Backend>>>`, without generic.

Generics make the codes more flexible, and make features such as static dispatch possible (so it can make codes run with different types at **compile time**), and profits from the advanced type system of Rust, it doesn't introduce extra mental burden (unlike C++). In fact, it's very tipical and appropriate to use generic in the case here: the same codes (in `StorageManager` and `S3Metadata`) work with different types (`Storage` implementors). This makes the `Storage`s like some kind of "plugins" (in compile time). 

So I think we should make full use of such a core mechanism, because flexible codes can respond to changing requirements.

xiaguan: I'm not saying generics are bad. My point might be about using associated types instead of generics. For example, `Iterator` uses associated types, not generics。

Reto911: The generic parameter of `S3Metadata` is removed. After `S3Backend`'s removal (may be in the next PR), `S3Metadata` and `S3Node` will be with no generic parameters.

xiaguan:
For instance, if the error is caused by network issues, degrading to write-through won't solve the problem either.

xiaguan:
After giving it some thought, storage shouldn't actually be placed in metadata, but rather in the file system (fs). Of course, decoupling metadata and storage is something we can address in future work.

Reto911:
> After giving it some thought, storage shouldn't actually be placed in metadata, but rather in the file system (fs). Of course, decoupling metadata and storage is something we can address in future work.

Yes, we should make it in the future. When this PR is merged, I will create an issue to track this.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -186,11 +201,52 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+                self.local_size.remove_with_guard(&ino, &guard).copied(),
+            )
+        };
+
         retry_txn!(TXN_RETRY_LIMIT, {
             let mut txn = self.kv_engine.new_meta_txn().await;
             let mut inode = self.get_inode_from_txn(txn.as_mut(), ino).await?;
-            inode.flush(ino, fh).await;
+
+            if inode.is_deferred_deletion() {
```
xiaguan, line 218: The flush should correspond to the close() operation, and I don't think it deletes the file. It looks like we need to investigate the maintenance of `open_count` and `lookup_count`.

Reto911: Yes, this is not to delete a file. But the deferred deletion is marked by `unlink()`, when `open_count` or `lookup_count` is **not** zero. Here checks the flag, and if it's true, there is no need to flush the file. This is following the old implementation of `flush()`.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -73,11 +80,16 @@ pub struct S3MetaData<S: S3BackEnd + Send + Sync + 'static> {
     pub(crate) kv_engine: Arc<KVEngineType>,
     /// Inum allocator
     inum_allocator: INumAllocator<KVEngineType>,
+    /// Mtime cache in local
+    local_mtime: HashMap<INum, SystemTime>,
```
xiaguan, line 84: Why don't we combine these two structs into a single tuple?  eg : HashMap<INum,(time,size)>

Reto911:
I agree with it, and the limited cache of mtime and size is just a makeshift, which must be removed when there is an advanced solution.

However, the storage system (especially the cache) depends on the local cached mtime, so I can't remove it from current PR, in current version of `master`.

src/async_fuse/memfs/s3_metadata.rs
Reto911: I will fix it later.

xiaguan (approved):

Issue #433, opened 2023-12-01
Removal of `etcd_delegate` and `dist` Modules 


xiaguan:
We propose to remove the `etcd_delegate` and `dist` modules from our codebase. The new `kv_engine` module effectively replaces their functionality, offering a more streamlined and clear approach.

- [ ] `etcd_delegate`: Redundant with `kv_engine`.
- [ ] `dist`: Currently not needed for distributed tasks. We can develop a new module if future needs arise, especially for p2p cache.

PR #466, opened 2024-01-25
cleanup: Remove deprecated modules

Reto911:
In this PR, we removed some modules that will never be used:
- Distributed cache, in `async_fuse/memfs/dist`, and the file `id_alloc.rs` is moved out of that dir. (#433)
- Old `S3Backend`, which is replaced by new `Backend` implemented with `openDAL`. By the way, the dependency `rust-s3` is also removed.
- Old `GlobalCache`, which is replaced by new `MemoryCache`.

Besides, we moved `async_fuse/memfs/cache` to the root of sources, and renamed it to `storage`.


src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -54,26 +48,19 @@ macro_rules! build_inconsistent_fs {
 const MY_TTL_SEC: u64 = 3600; // TODO: should be a long value, say 1 hour
 /// The generation ID of FUSE attributes
 const MY_GENERATION: u64 = 1; // TODO: find a proper way to set generation
-#[allow(dead_code)]
 /// The limit of transaction commit retrying times.
 const TXN_RETRY_LIMIT: u32 = 10;
 
 /// File system in-memory meta-data
 #[derive(Debug)]
 #[allow(dead_code)]
```
xiaguan, line 56: Is this mark still available?

src/async_fuse/memfs/s3_node.rs
```diff
@@ -45,7 +40,7 @@ pub enum S3NodeData {
     /// Directory entry data
     Directory(BTreeMap<String, DirEntry>),
     /// File content data
-    RegFile(Arc<GlobalCache>),
+    RegFile,
```
xiaguan, line 43: After dir entry refactor, the `NodeData` seems meaning less :thinking:

src/async_fuse/memfs/s3_node.rs
```diff
@@ -62,17 +57,15 @@ impl S3NodeData {
...
         }
     }
 }
 
 /// A file node or a directory node
 #[derive(Debug)]
-pub struct S3Node<S: S3BackEnd + Sync + Send + 'static> {
-    /// S3 Backend
-    s3_backend: Arc<S>,
+pub struct S3Node {
```
xiaguan, line 68: We should rename `S3Metadata` and `S3Node` since they have nothing to do with "S3".

xiaguan:
This pull request is a simple code structure refactor.  LGTM.

src/async_fuse/memfs/s3_node.rs
Reto911: Yes, the only thing it holds is the target path of symlink.

src/async_fuse/memfs/s3_node.rs
Reto911: Make sense. And the trait of `Metadata` and `Node` should be considered further.

src/async_fuse/memfs/s3_metadata.rs
Reto911: Yes, the field named `cur_fd` is never used. I don't know what is it.

xiaguan (approved):
LGTM

Issue #456, opened 2024-01-08
Implementation of the file deletion logic


xiaguan:
After the storage refactoring is complete, we should implement our file deletion logic.

- [x] Implementation of the file deletion logic

Reto911:
#463

Reto911:
Overwritten in #467, needs to be fixed.

PR #469, opened 2024-01-29
feat(metadata): add file's open-to-close support

xiaguan:
In this pull request, I've added an open-to-close metadata cache for files. This significantly enhances read and write performance. There's an issue worth noting, though. When flushing to the KV store on close, do we need to call set_attr() to overwrite all remote information, or should we just update the IO-related attributes like time and size?

### DIFF
--- a/src/async_fuse/memfs/open_file.rs
+++ b/src/async_fuse/memfs/open_file.rs
@@ -69,7 +69,6 @@ impl OpenFiles {
     /// Retrieves an `OpenFile` by its inode number.
     ///
     /// Panics if the file is not found.
-    #[allow(clippy::unwrap_used)] // We should panic here
     pub fn get(&self, inum: INum) -> OpenFile {
         self.try_get(inum)
             .unwrap_or_else(|| panic!("Couldn't find the open file with inum={inum}"))
@@ -80,9 +79,15 @@ impl OpenFiles {
     /// Returns a reference to the newly opened file.
     pub fn open(&self, inum: INum, attr: FileAttr) -> OpenFile {
         let mut inner = self.inner.lock();
-        let open_file = Arc::new(RwLock::new(RawOpenFile { attr, open_cnt: 1 }));
-        inner.open_files.insert(inum, Arc::clone(&open_file));
-        open_file
+        let open_file = inner
+            .open_files
+            .entry(inum)
+            .or_insert_with(|| Arc::new(RwLock::new(RawOpenFile { attr, open_cnt: 0 })));
+        {
+            let mut open_file = open_file.write();
+            open_file.open_cnt += 1;
+        }
+        Arc::clone(open_file)
     }
 
     /// Tries to open a file if it is already in the collection.
@@ -105,7 +110,6 @@ impl OpenFiles {
     /// if its open count reaches zero. Returns the `OpenFile` if it was
     /// removed, or `None` if it remains open.
     #[allow(clippy::unwrap_in_result)]
-    #[allow(clippy::unwrap_used)]
     pub fn close(&self, inum: INum) -> Option<OpenFile> {
         let mut inner = self.inner.lock();
         let open_count = {
--- a/src/async_fuse/memfs/open_file.rs
+++ b/src/async_fuse/memfs/open_file.rs
@@ -1,0 +1,127 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use parking_lot::{Mutex, RwLock};
+
+use super::fs_util::FileAttr;
+use crate::async_fuse::fuse::protocol::INum;
+
+/// A structure representing an open file with its attributes and open count.
+///
+/// The `attr` field contains the file attributes, while `open_cnt` keeps track
+/// of the number of times this file is currently opened.
+#[derive(Debug)]
+#[allow(clippy::partial_pub_fields)] // Pub attr for simplicity.
+pub struct RawOpenFile {
+    /// The file attributes.
+    pub attr: FileAttr,
+    /// The number of times this file is currently opened.
+    open_cnt: u32,
+}
+
+/// A thread-safe reference counted wrapper around `RawOpenFile`.
+///
+/// This type uses `Arc` for shared ownership and `RwLock` for thread-safe
+/// mutability, allowing multiple readers or one writer at the same time.
+pub type OpenFile = Arc<RwLock<RawOpenFile>>;
+
+/// Internal structure to keep track of all open files.
+///
+/// This structure uses a `HashMap` to associate inode numbers (`INum`) with
+/// their corresponding `OpenFile` instances.
+#[derive(Debug)]
+struct RawOpenFiles {
+    /// The collection of open files.
+    open_files: HashMap<INum, OpenFile>,
+}
+
+/// A thread-safe, mutable collection of open files.
+///
+/// This structure provides an interface to interact with open files,
+/// encapsulating the internal locking mechanism through `Mutex`.
+#[derive(Debug)]
+pub struct OpenFiles {
+    /// The internal collection of open files.
+    inner: Arc<Mutex<RawOpenFiles>>,
+}
+
+impl OpenFiles {
+    /// Constructs a new `OpenFiles` collection.
+    ///
+    /// This initializes an empty collection of open files.
+    pub fn new() -> Self {
+        OpenFiles {
+            inner: Arc::new(Mutex::new(RawOpenFiles {
+                open_files: HashMap::new(),
+            })),
+        }
+    }
+
+    /// Try to retrieves an `OpenFile` by its inode number.
+    ///
+    /// Returns `Some(OpenFile)` if the file is open, or `None` if not found.
+    pub fn try_get(&self, inum: INum) -> Option<OpenFile> {
+        let inner = self.inner.lock();
+        let open_file = inner.open_files.get(&inum).map(Arc::clone);
+        open_file
+    }
+
+    /// Retrieves an `OpenFile` by its inode number.
+    ///
+    /// Panics if the file is not found.
+    #[allow(clippy::unwrap_used)] // We should panic here
+    pub fn get(&self, inum: INum) -> OpenFile {
+        self.try_get(inum)
+            .unwrap_or_else(|| panic!("Couldn't find the open file with inum={inum}"))
+    }
+
+    /// Opens a file, adding it to the collection
+    ///
+    /// Returns a reference to the newly opened file.
+    pub fn open(&self, inum: INum, attr: FileAttr) -> OpenFile {
+        let mut inner = self.inner.lock();
+        let open_file = Arc::new(RwLock::new(RawOpenFile { attr, open_cnt: 1 }));
+        inner.open_files.insert(inum, Arc::clone(&open_file));
+        open_file
+    }
+
+    /// Tries to open a file if it is already in the collection.
+    ///
+    /// Increments the open count if the file is found. Returns `Some(OpenFile)`
+    /// if successful, or `None` if the file is not in the collection.
+    pub fn try_open(&self, inum: INum) -> Option<OpenFile> {
+        let mut inner = self.inner.lock();
+        let open_file = inner.open_files.get_mut(&inum)?;
+        {
+            let mut open_file = open_file.write();
+            open_file.open_cnt += 1;
+        }
+        Some(Arc::clone(open_file))
+    }
+
+    /// Closes an open file, identified by its inode number.
+    ///
+    /// Decrements the open count and removes the file from the collection
+    /// if its open count reaches zero. Returns the `OpenFile` if it was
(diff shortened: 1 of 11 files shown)


src/async_fuse/memfs/open_file.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Opens a file, adding it to the collection
+    ///
+    /// Returns a reference to the newly opened file.
+    pub fn open(&self, inum: INum, attr: FileAttr) -> OpenFile {
+        let mut inner = self.inner.lock();
+        let open_file = Arc::new(RwLock::new(RawOpenFile { attr, open_cnt: 1 }));
+        inner.open_files.insert(inum, Arc::clone(&open_file));
+        open_file
+    }
```
Reto911, line 86: Do we need to maintain some attributes when a file is opened again? For example, if something changed from the remote nodes and they are flushed to the KV, would them be visible for the local node?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -305,28 +269,49 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+        // The file was not opened before, so we need to open it.
+        let (result, attr) = retry_txn!(TXN_RETRY_LIMIT, {
             let mut txn = self.kv_engine.new_meta_txn().await;
             let node = self.get_inode_from_txn(txn.as_mut(), ino).await?;
-            let o_flags = fs_util::parse_oflag(flags);
-            node.open_pre_check(o_flags, context.user_id, context.group_id)?;
-
+            let attr = node.get_attr();
+            attr.check_perm(context.user_id, context.group_id, access_mode)?;
             let result = node.dup_fd(o_flags).await;
```
Reto911, line 292: If we do not maintain the `open_count` in `S3Node` really, we can remove the field from `S3Node`, so that an `open()` call would not need to commit anything to the KV.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -365,69 +350,20 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
-                        }
+            let applied_attr = origin_attr;
+
+            let dirty_attr_for_reply =
+                match applied_attr.setattr_precheck(param, context.user_id, context.group_id)? {
+                    Some(dirty_attr) => {
+                        let dirty_attr_for_kv = dirty_attr;
+                        inode.set_attr(dirty_attr_for_kv);
+                        dirty_attr
                     }
```
Reto911, line 361: Why do you remove the calling of `truncate`?

Reto911: Storage related attributes (`mtime` and `size`) will be committed immediately when `setattr` is called, is it right?

Reto911: And, how about `open_files`? Why don't you maintain it in this function?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -504,7 +440,6 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
                 );
             } else {
                 if let SFlag::S_IFREG = child_node.get_type() {
-                    self.local_mtime_and_size.remove(&child_ino);
```
Reto911, line 507: We maintain `open_count` in `open_files` now, so the `deferred_deletion` should refer to `open_files` but not `S3Node::open_count`, right?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -617,7 +552,6 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
             if let SFlag::S_IFREG = node.get_type() {
                 let ino = node.get_ino();
                 self.storage.remove(ino).await;
-                self.local_mtime_and_size.remove(&ino);
```
Reto911, line 620: We don't maintain `open_files` in delete related functions, because the open file entry must be removed when `release()` called, right?

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -718,7 +652,7 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
 
             let child_ino = child_entry.ino();
             let child_node = self.get_inode_from_txn(txn.as_mut(), child_ino).await?;
-            let child_attr = self.apply_local_mtime_and_size(child_node.lookup_attr());
+            let child_attr = child_node.get_attr();
```
Reto911, line 655: Shall we look up to `open_files` for the attributes when the file is opened?

src/async_fuse/memfs/cache/storage_manager.rs
```diff
@@ -44,20 +44,9 @@ where
...
-        }
-
         let mut blocks = vec![];
-
-        for (handle, block_id) in handles.into_iter().zip(start_block..end_block) {
-            let block = handle
-                .await
-                .unwrap_or_else(|e| panic!("Fails when awaiting on loading: {e}"));
+        for block_id in start_block..end_block {
+            let block = self.storage.load(ino, block_id).await;
```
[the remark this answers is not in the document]
xiaguan: Yes

src/async_fuse/memfs/open_file.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn try_get(&self, inum: INum) -> Option<OpenFile> {
+        let inner = self.inner.lock();
+        let open_file = inner.open_files.get(&inum).map(Arc::clone);
+        open_file
+    }
+
+    /// Retrieves an `OpenFile` by its inode number.
+    ///
+    /// Panics if the file is not found.
+    #[allow(clippy::unwrap_used)] // We should panic here
```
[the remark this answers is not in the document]
xiaguan: Yes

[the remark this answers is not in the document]
Reto911: But I tried it locally, and the clippy checks pass🤔
It seems that `unwrap_or_else` is OK, and we've used it every where already.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -98,16 +98,26 @@ impl<S: S3BackEnd + Sync + Send + 'static> MetaData for S3MetaData<S> {
...
+                node.set_attr(attr);
+                txn.set(
+                    &KeyType::INum2Node(ino),
+                    &ValueType::Node(node.to_serial_node()),
+                );
+                (txn.commit().await, ())
+            })?;
+        } else {
+            info!("release() ino={} fh={} file is still open", ino, fh);
+        }
```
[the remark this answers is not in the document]
xiaguan: I have redesigned the following. Metadata now does not require refreshing. Still WIP.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: Removed

src/async_fuse/memfs/s3_metadata.rs
xiaguan: open-to-close only cares about IO operations (expect the `getattr()` ). Other opeartions just directly use the kv.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: It just refer to `lookup_cnt` only.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: No

src/async_fuse/memfs/s3_metadata.rs
xiaguan: `lookup_cnt == 0` means no kernel still use the node. In other words, the file is not open.

xiaguan:
The `write_helper` of the metadata will immediately take effect on the kv, updating its mtime and size. This way, we won't need to worry too much about refreshing metadata frequently. Of course, we need to separate the `Storage::store` and `write_helper` to achieve better performance

src/async_fuse/memfs/s3_metadata.rs
Reto911: > open-to-close only cares about IO operations (expect the `getattr()` ). Other opeartions just directly use the kv.

So, we must call `truncate` here because it's an IO operation. And `mtime` and `size` should be updated to `open_files` for the following `read` and `write` calls.

src/async_fuse/memfs/s3_metadata.rs
Reto911: `node` seems not to be modified, so do we still need to commit the node to KV?

src/async_fuse/memfs/open_file.rs
xiaguan: No , they just see the local node

src/async_fuse/memfs/s3_metadata.rs
xiaguan: I will fix it in this pull request later.

src/async_fuse/memfs/s3_metadata.rs
xiaguan: fixed.

Reto911: So the `setattr` will never interact with `open_files`, would it?

xiaguan: Yes

Reto911:
If we commit to KV immediately when `write()`, and update the atime to KV often when `read()`, then what's the meaning of `open-to-close`?

src/async_fuse/memfs/open_file.rs
```diff
@@ -0,0 +1,131 @@
...
+                .open_files
+                .get_mut(&inum)
+                .unwrap_or_else(|| panic!("Couldn't find the open file with inum={inum}"));
+            let mut open_file = open_file.write();
+            debug_assert!(open_file.open_cnt > 0);
+            open_file.open_cnt -= 1;
+            open_file.open_cnt
+        };
+        if open_count == 0 {
+            inner.open_files.remove(&inum)
```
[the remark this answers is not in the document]
xiaguan: When get open_count , we already make sure it is exist. SInce we hold inner's lock.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -87,16 +88,12 @@ impl MetaData for S3MetaData {
...
-                &ValueType::Node(inode.to_serial_node()),
-            );
-            (txn.commit().await, ())
-        })?;
+        if self.open_files.close(ino).is_some() {
+            // open_count reaches 0, flush the metadata to kv
+            info!("release() ino={} fh={} file is closed", ino, fh);
+        } else {
+            info!("release() ino={} fh={} file is still open", ino, fh);
+        }
```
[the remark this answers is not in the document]
xiaguan: Currently is no need to flush.

src/async_fuse/memfs/s3_metadata.rs
```diff
@@ -327,8 +315,13 @@ impl MetaData for S3MetaData {
...
             let inode = self.get_inode_from_txn(txn.as_mut(), ino).await?;
             inode.dec_lookup_count_by(nlookup);
-            let is_deleted = self.delete_check(&inode).await?;
+            let is_deleted = inode.get_lookup_count() == 0;
             if is_deleted {
+                // FIXME: rename should also rename the node's name and reset the parent ino
+                txn.delete(&KeyType::DirEntryKey((
+                    inode.get_parent_ino(),
+                    inode.get_name().to_owned(),
+                )));
```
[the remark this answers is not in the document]
xiaguan: Fixed.

xiaguan:
> If we commit to KV immediately when `write()`, and update the atime to KV often when `read()`, then what's the meaning of `open-to-close`?

Closing a file forces dirty data and metadata to be flushed to the server, and opening a file forces the client to contact the server in order to revalidate cached data.

When a file was closed, reopen it anywhere anytime should see the changes from last change.
We did follow this rule.

src/async_fuse/memfs/open_file.rs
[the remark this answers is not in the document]
Reto911: I know, this is guarenteed. This is not a problem, just making the code clearer.

src/async_fuse/memfs/s3_metadata.rs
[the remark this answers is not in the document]
Reto911: And you are not going to fix #456 in this PR, right?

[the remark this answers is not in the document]
xiaguan: Yes.

Reto911 (approved):

PR #473, opened 2024-01-30
fix: handle errors in `Storage`

Reto911:
1. Make `WriteBackTask` into a struct instead of a single async fn.
2. Splict `storage_manager.rs` into seperated files (submodules).
3. Functions in `storage` mod returns `Error` instead of panic. 

A single exception of *3* above is `MemoryCache::send_block_to_write_back_task`: 
```rust
panic!("Command receiver in write back task is closed unexpectedly.")
```
This will be guaranteed to be unreachable soon.

### DIFF
--- a/src/storage/backend/backend_impl.rs
+++ b/src/storage/backend/backend_impl.rs
@@ -8,8 +8,8 @@ use opendal::services::{Fs, S3};
 use opendal::{ErrorKind, Operator};
 
 use crate::async_fuse::fuse::protocol::INum;
-use crate::storage::error::{StorageError, StorageOperation, StorageResult};
-use crate::storage::{Block, Storage};
+use crate::storage::error::StorageResult;
+use crate::storage::{Block, Storage, StorageError, StorageOperation};
 
 /// Get file path by `ino`
 fn get_file_path(ino: INum) -> String {
--- a/src/storage/backend/backend_impl.rs
+++ b/src/storage/backend/backend_impl.rs
@@ -9,7 +9,7 @@ use opendal::{ErrorKind, Operator};
 
 use crate::async_fuse::fuse::protocol::INum;
 use crate::storage::error::StorageResult;
-use crate::storage::{Block, Storage, StorageError, StorageOperation};
+use crate::storage::{Block, Storage};
 
 /// Get file path by `ino`
 fn get_file_path(ino: INum) -> String {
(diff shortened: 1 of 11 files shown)


Reto911:
I will double-check the coverage issues.

Reto911:
Waiting for #476

src/storage/backend/backend_impl.rs
```diff
@@ -178,9 +200,10 @@ impl Storage for Backend {
         self.operator
             .remove_all(&get_file_path(ino))
             .await
-            .unwrap_or_else(|e| {
-                panic!("Failed to remove a file from the backend of ino={ino}: {e}")
-            });
+            .map_err(|e| StorageError {
```
xiaguan, line 203: Use `?` instead of `map_err`.

Reto911: `e` cannot be converted into `StorageError` directly, because we must fill the field `operation` manually.

xiaguan: I feel like there's no need to store operation info in the error. If we need that info, we could just let the caller use `error!` to log it. The thing is, `map_err` is kind of spreading all over the code, and it doesn’t look too pretty.

Reto911: Ok, I will fix this later.

xiaguan: Maybe we need a `BackEndError` ?

Reto911: > Maybe we need a `BackEndError` ?

What does it contain? I think `StorageErrorInner` (which will be renamed to `StorageError`) is enough, it contains an opendal error, with the context of the backend operation.

xiaguan: Since memory cache is currently error free. The StorageErorr is equal to opendal::error ? If so ,just use the opendal::error. 
But when disk cache is supported, cache will gen error also，which might be `std::io::errror`.

Reto911: > Since memory cache is currently error free. The StorageErorr is equal to opendal::error ? If so ,just use the opendal::error. But when disk cache is supported, cache will gen error also，which might be `std::io::errror`.

opendal operations will also return `std::io::Error`, for example, in `Writer`s and `Reader`s. Therefore, the enum of:

- `std::io::Error`
- `opendal::Error`
- `OutOfRange` (for blocks)
- `Internal` (anyhow errors)

is enough and not redundent. I will use this enum directly, so that we can use `?` here.

xiaguan (approved):

PR #476, opened 2024-01-31
refactor(storage): make big files into submodules

Reto911:
In this PR, we refactored some big single-file modules into several small submodules, making the codes clearer and easier to be maintained.

Affected modules are listed below:
1. `storage_manager`, splicted into small submodules, without other changes.
2. `write_back_task`, make `WriteBackTask` into a struct instead of a function. Besides, it holds a `Arc` instead of `Weak` now.
3. `backend`, splicted into small submodules, without other changes.


xiaguan (approved):
It seems to be a structural refactor. I didn't go through it line by line. LGTM.

PR #477, opened 2024-02-02
feat: Introduce Metrics

Reto911:
In this PR, we introduced `metrics` modules for DatenLord, with metrics listed below:

- `cache`
- `file_system`
- `kv`

The metrics for `Backend` are delegated to `opendal::layers::PrometheusLayer` for now.

A trait called `LossyCast` is introduced, for casting from integer types to floating types with potentially lossy precision.

Refactored `retry_txn` macro to return the retry times.


frostyplanet:
Is there option to turn off latency metrics statistics?

Reto911:
No. Perhaps we can add one later.

frostyplanet (approved):

PR #479, opened 2024-02-18
chore: add `rust-toolchain` file

Reto911:


Reto911:
This is part of #480 now.

PR #480, opened 2024-02-19
chore: add `rust-toolchain.toml` and `devcontainer.json`

Reto911:
In this PR, we added `rust-toolchain.toml` file to make it easier to set up the develop environment.

In addition, we added a `devcontainer.json` file to help setting up a dev container for developers to quick start. (Thanks to @yunwei37 's contribution, which is shown in #418 ).

The `rustup-init.sh` script is just downloaded from <https://sh.rustup.rs>.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+{
+    "name": "DatenLord Dev Container",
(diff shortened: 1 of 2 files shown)


xiaguan:
Can we use a DockerFile to specify the image we need, in order to avoid directly copying `rustup-init.sh` ?

Reto911:
I will try it later.

xiaguan (approved):

Issue #449, opened 2023-12-31
replace `user_id` and `group_id` with `uid` and `gid`


xiaguan:
I would like to propose a modification in our codebase to replace the terms "user_id" and "group_id" with their corresponding abbreviations, "uid" and "gid". This change aims to enhance code readability and conciseness.

It's very easy, just a naming issue.

rudransh61:
Ok i will do it

rudransh61:
Can you tell me the files in which they are used  , or i will manually find them

xiaguan:
All of them. Just manually find them.

rudransh61:
https://github.com/datenlord/datenlord/pull/451
Done pls review it

PR #481, opened 2024-02-19
chore: Rename variables.

IRONICBo:
In this PR, we fixed the issue #449 to enhance code readability and conciseness.

### DIFF
--- a/src/storage/backend/backend_impl.rs
+++ b/src/storage/backend/backend_impl.rs
@@ -9,7 +9,7 @@ use opendal::{ErrorKind, Operator};
 
 use crate::async_fuse::fuse::protocol::INum;
 use crate::storage::error::StorageResult;
-use crate::storage::{Block, Storage, StorageError, StorageOperation};
+use crate::storage::{Block, Storage};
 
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -69,6 +69,7 @@
     clippy::impl_trait_in_params,  // Allow impl AsRef<Path>, it's common in Rust
     clippy::missing_assert_message, // Allow assert! without message, mainly in test code
     clippy::semicolon_outside_block, // We need to choose between this and `semicolon_inside_block`, we choose outside
+    clippy::similar_names, // Allow similar names
```
xiaguan, line 72: Provide specific reasons, due to the existence of uid and gid

xiaguan:
1. The priority of using rebase is higher than merge
2. The commit message should be clearer, specifying which variables have been renamed

src/main.rs
IRONICBo: Ok, I will update it.

xiaguan:
Update the branch and rerun failed tests.

Issue #437, opened 2023-12-04
Enhance Configuration Options with File and Environment Variable Support


xiaguan:
## Feature Request

**Implementation Details:**

1. **Configuration File Support:**
   - Integrate the ability to read configurations from a file.
   - Require enabling serialization features to facilitate this functionality.
   - Support common file formats (e.g., JSON, TOML) for easy integration and use.

2. **Environment Variable Support:**
   - Implement the capability to set configurations via environment variables.
   - This approach should be similar to the file-based configuration, leveraging serialization where appropriate.

**Configuration Priority:**
With the introduction of these new methods, the priority for configuration settings will be as follows:
   1. **Command-Line Arguments:** Highest priority. Any settings provided here will override those from other sources.
   2. **Environment Variables:** Intermediate priority. These will take precedence over file-based configurations but can be overridden by command-line arguments.
   3. **Configuration File:** Lowest priority. These settings provide a default configuration but can be overridden by environment variables or command-line arguments.

**Expected Outcome:**
By supporting these additional configuration methods, we aim to offer users more versatility in how they configure the application. This will be particularly useful in different deployment scenarios, such as containerized environments or cloud deployments, where flexibility in configuration is key.

**Request for Input:**
We welcome suggestions and discussions on the best practices for implementing these features, especially regarding the choice of file formats and any potential challenges in integrating these configuration methods seamlessly with our existing system.

PR #483, opened 2024-02-21
feat: Enhance config options with toml file and env variables

IRONICBo:
This PR use `serfig` to load config from `file`, `env`, and `command line`, then build then with priority(file < env < args).

- Reference issue: https://github.com/datenlord/datenlord/issues/437

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ ring-io = { git = "https://github.com/datenlord/ring-io", rev = "2f0506c" }
 serde-xml-rs = "0.6"
 serde = "1.0.126"
 serde_json = "1.0.64"
+toml = "0.7.3"
 tar = "0.4.29"
 thiserror = "1.0.22"
 tiny_http = "0.10.0"
@@ -67,6 +68,7 @@ opendal = {version = "0.43.0", features = ["layers-prometheus"]}
 signal-hook = { version = "0.3.17", features = ["iterator"]}
 signal-hook-tokio = { version = "0.3.1", features = ["futures-v0_3"] }
 tokio-util = "0.7.10"
+serfig = "0.1.0"
 
 [build-dependencies]
 protoc-grpcio = "3.0.0"
@@ -76,6 +78,7 @@ mock-etcd = { git = "https://github.com/datenlord/etcd-client", rev = "f697899" 
 #  Create mock api for some interface like s3 that are independent service,
 #  We don't want to start up a real service for some api when doing unit test.
 mockall = "0.11.2"
+tempfile = "3"
 
 [[bin]]
 path = "src/bin/bind_mounter.rs"
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -1,24 +1,34 @@
 use clap::Parser;
+use serde::{Deserialize, Serialize};
+use serfig::collectors::{from_env, from_file, from_self};
+use serfig::parsers::Toml;
 
-#[derive(Debug, Parser)]
+use crate::common::error::DatenLordError;
+
+#[derive(Debug, Clone, Serialize, Deserialize, Parser, Default)]
 #[clap(author, version, about, long_about = None)]
+#[serde(default)]
 /// A config
 pub struct Config {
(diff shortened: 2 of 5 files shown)


src/config/config.rs
```diff
@@ -139,17 +157,185 @@ pub struct CSIConfig {
...
+    /// Check the config values are valid
+    #[inline]
+    pub fn check_valid(&self) -> Result<bool, DatenLordError> {
+        // Check role/node_name/node_ip/mount_path
+        let checked_values = [&self.role, &self.node_name, &self.node_ip, &self.mount_path];
+
+        let valid = checked_values.iter().all(|x| x.is_some());
+
+        Ok(valid)
+    }
```
Reto911, line 215: It seems that this function does not return any error, why do you make it return a `Result`?

Cargo.toml
```diff
@@ -65,6 +66,8 @@ tracing-subscriber = "0.3"
 hashlink = "0.8.4"
 clap = { version = "4", features = ["derive"] }
 opendal = "0.43.0"
+serfig = "0.1.0"
+tempfile = "3"
```
Reto911, line 70: If it's only shown in test codes, just move it under "dev-dependencies"

src/config/config.rs
```diff
@@ -139,17 +157,185 @@ pub struct CSIConfig {
...
+            kv_server_list = [
+              "127.0.0.1:2379"
+            ]
+        "#
+        )
+        .unwrap();
+
+        let mut builder: serfig::Builder<Config> = serfig::Builder::default();
+        builder = builder.collect(from_file(Toml, temp_file_path.to_str().unwrap()));
+        let conf = builder.build().unwrap();
```
Reto911, line 254: You seems to be writing UT for `serfig` crate, but not for `datenlord`. Perhaps you should refactor `Config::load()` to make it possible to test your own logic. In fact, the `Config::load()` method is not coverred by test cases at all.

src/config/config.rs
IRONICBo: I will update it, it was initially designed to construct an exception, but later it turned out not to be very useful

src/config/config.rs
IRONICBo: Ok, I try to load config from basic `config::parse()`, and it can be integrated with unit test.

Some mock operation may be not elegant in unit test.
> https://www.reddit.com/r/rust/comments/xig8xm/best_method_for_cli_testing_with_clap/

src/config/config.rs
```diff
@@ -139,17 +157,185 @@ pub struct CSIConfig {
...
+        // Load from arguments
+        builder = builder.collect(from_self(arg_conf));
+
+        // Build the config with multi sources
+        let conf = builder.build()?;
+
+        // Check valid
+        match conf.check_valid() {
+            Ok(true) => Ok(conf),
+            Ok(false) => Err(DatenLordError::ArgumentInvalid { context: vec![] }),
```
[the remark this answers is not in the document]
Reto911: You haven't print the debug-formatted `conf` into the context yet.

```rust
Err(DatenLordError::ArgumentInvalid {
    context: vec!["Current command line arguments are not valid.".to_owned(), format!("{conf:?}")],
})
```

Reto911:
By the way, please resolve the conflicts at your next commit.

frostyplanet (dismissed):

src/config/config.rs
```diff
@@ -139,17 +153,178 @@ pub struct CSIConfig {
...
+
+        // Build the config with multi sources
+        let conf = builder.build()?;
+
+        // Check valid
+        if conf.check_valid() {
+            Ok(conf)
+        } else {
+            Err(DatenLordError::ArgumentInvalid {
+                context: vec!["Current command line arguments are not valid.".to_owned()],
```
rogercloud, line 187: As you've checked the configuration's valication first, you'd better tell the user which parameter is not right, which is a helpful fix hint.

IRONICBo: Thanks, I will change it.

frostyplanet (approved):

Issue #482, opened 2024-02-20
Wrong usage of Reader::read


xiaguan:
https://github.com/datenlord/datenlord/blob/aecad64a1a91e5dc01ef9d4c730705c6b281d406/src/storage/backend/backend_impl.rs#L105

According to this ,https://github.com/xiaguan/page_cache/issues/1

AuruTus:
Hi, I have interest in this. Can assign it to me?

xiaguan:
@AuruTus  It was solved by #484 .

xiaguan:
@Reto911 will later release some good first issues of fuse. You can keep an eye out for them. If you plan to participate in GSoC, you can directly work on the #387

PR #484, opened 2024-02-21
fix: check returning read size

IRONICBo:
Try to check if the `Reader::read` has reached `0`.

Reference issue: https://github.com/datenlord/datenlord/issues/482

### DIFF
--- a/src/storage/backend/backend_impl.rs
+++ b/src/storage/backend/backend_impl.rs
@@ -110,21 +110,39 @@ impl Storage for Backend {
     async fn load_from_self(&self, ino: INum, block_id: usize) -> StorageResult<Option<Block>> {
         let mut block = Block::new_zeroed(self.block_size);
 
-        if let Err(e) = self
-            .operator
-            .reader(&get_block_path(ino, block_id))
-            .await?
-            .read(block.make_mut_slice())
-            .await
-        {
-            if e.kind() == std::io::ErrorKind::NotFound {
-                Ok(None)
-            } else {
-                Err(e.into())
+        let mut reader = self.operator.reader(&get_block_path(ino, block_id)).await?;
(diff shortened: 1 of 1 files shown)


xiaguan:
Regarding this bug, we should focus more on why no tests were able to cover this case.

src/storage/backend/backend_impl.rs
```diff
@@ -98,21 +98,29 @@ impl Storage for Backend {
...
-            .await
-        {
-            if e.kind() == std::io::ErrorKind::NotFound {
-                Ok(None)
-            } else {
-                Err(e.into())
+        let mut reader = self.operator.reader(&get_block_path(ino, block_id)).await?;
+        // Check if the reader point is at the end of the file.
+        loop {
+            match reader.read(block.make_mut_slice()).await {
```
xiaguan, line 104: call `make_mut_slice()`  outside the loop ,it is not zero cost

xiaguan (changes requested):

Reto911:
> Regarding this bug, we should focus more on why no tests were able to cover this case.

In another project of mine, I found that `File` can read 2 MiB at most in each `read()` call, I think these are similar issues. In this project, the block size is 512 KiB, which does not reach the limit.

src/storage/backend/backend_impl.rs
Reto911: The block is just created in this method, and the reference count of it should always be 1, therefore `make_mut_slice` should not perform any copy, and use it in the loop is OK.

However, the start point of the block slice for `reader.read` to write is always 0 in this case, and when the loop continues, it will overwrite the contents which are readed in the last iteration. You should set a varible to record the start point in each iteration.

xiaguan (approved):

PR #486, opened 2024-02-23
chore: rename user_id and group_id to uid and gid.

IRONICBo:
In this PR, we fixed the issue #449 to enhance code readability and conciseness.

### DIFF
--- a/src/async_fuse/memfs/fs_util.rs
+++ b/src/async_fuse/memfs/fs_util.rs
@@ -73,8 +73,8 @@ impl FileAttr {
     pub(crate) fn setattr_precheck(
         &self,
         param: &SetAttrParam,
-        user_id: u32,
-        group_id: u32,
+        context_uid: u32,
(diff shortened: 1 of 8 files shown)


src/async_fuse/fuse/mount.rs
```diff
@@ -166,7 +166,7 @@ async fn direct_mount(mount_point: &Path) -> anyhow::Result<RawFd> {
             "failed to get the file stat of mount point={mount_point:?}",
         ))?;
     let opts = format!(
-        "fd={},rootmode={:o},user_id={},group_id={}",
+        "fd={},rootmode={:o},uid={},gid={}",
```
Reto911, line 169: You cannot rename these, for these are mount arguments for syscall `mount`

IRONICBo: ok, I will update these.

Reto911 (approved):

Issue #495, opened 2024-03-12
To Asynchronize and Manage the `fuse_fd` 


Reto911:
It is necessary to encapsulate the fd of the FUSE device, clearly manage its lifetime and ownership, and minimize direct interaction with the fd as much as possible to ensure [IO safety](https://doc.rust-lang.org/std/io/index.html#io-safety). Additionally, implement an fd pool or a similar mechanism to avoid always reusing the same Session fd, thereby preventing performance losses due to kernel locking in high-concurrency scenarios.

There are three targets of this issue:

1. **fd Management**: Currently, we communicate with the FUSE kernel module by directly invoking system calls `read` and `writev` on the raw `fuse_fd`. For safety reasons, we should further encapsulate the fd to avoid direct interaction with the raw fd. This encapsulation should be implemented in an RAII style to clarify ownership and lifetime. This is also the issue addressed in #165. Currently, we have implemented the FuseFd structure for automatic release of the `fuse_fd`, but we still rely on raw system calls and have not avoided direct interaction with the fd.
2. **Asynchronization**: We invoke `read` and `writev` synchronously in-place during FUSE message reception and transmission now. Since these system calls are blocking, we wrap them with spawn_blocking. It is necessary for us to implement a more efficient asynchronous solution for reading and writing to the fuse_fd. Therefore, the encapsulation of the raw fd mentioned in the previous point should choose a solution that facilitates asynchronous implementation.
3. **fd Pool**: According to #10 and #28, we should use `ioctl` in multi-threaded mode to allocate a separate `Worker fd` for each worker thread instead of always reusing the Session fd to avoid the performance bottleneck caused by kernel locking. This is also the role of the currently unused channel module. Of course, these separate `Worker fd`s should also be encapsulated according to the principles mentioned in the previous two points.

Besides, we can use `std::io::IoSlice` instead of `IoVec` in `nix`.

Issue #28, opened 2020-07-01
FUSE device file handler pool


pwang7:
When processing FUSE requests concurrently, it's better to not share one FUSE device file handler but have multiple ones, so as to avoid [the underlying lock associated with FUSE device.](https://john-millikin.com/the-fuse-protocol#multi-threading).

So it might need to pre-open multiple FUSE device file handlers, and put them inside a pool to share.

Issue #10, opened 2020-06-27
Use channel fd instead of session fd in multi-thread mode


rogercloud:
For performance issue, we clone the device fd for each channel, one channel per Task, to avoid fd lock bottleneck.

PR #505, opened 2024-03-19
feat: make FUSE device IO asynchronous

Reto911:
In this PR, we avoided to use raw FUSE fd to interact with FUSE kernel module, by wrapping the raw fd with `std::fs::File`. And separated threads are used to receive FUSE requests, for avoiding invoking `spawn_blocking` always.

`IoVec` is deprecated, we use `std::io::IoSlice` now.

Besides, in this PR, we use `worker fd` for each FUSE request handlers, reducing potential resources race.

### DIFF
--- a/src/async_fuse/fuse/file_system.rs
+++ b/src/async_fuse/fuse/file_system.rs
@@ -31,32 +31,32 @@ pub trait FileSystem {
         req: &Request<'_>,
         parent: INum,
         name: &str,
-        reply: ReplyEntry,
+        reply: ReplyEntry<'_>,
     ) -> nix::Result<usize>;
 
     /// Forget about an inode
     async fn forget(&self, req: &Request<'_>, nlookup: u64);
 
     /// Get file attributes.
-    async fn getattr(&self, req: &Request<'_>, reply: ReplyAttr) -> nix::Result<usize>;
+    async fn getattr(&self, req: &Request<'_>, reply: ReplyAttr<'_>) -> nix::Result<usize>;
 
     /// Set file attributes.
     async fn setattr(
         &self,
         req: &Request<'_>,
         param: SetAttrParam,
-        reply: ReplyAttr,
+        reply: ReplyAttr<'_>,
     ) -> nix::Result<usize>;
 
     /// Read symbolic link.
-    async fn readlink(&self, req: &Request<'_>, reply: ReplyData) -> nix::Result<usize>;
+    async fn readlink(&self, req: &Request<'_>, reply: ReplyData<'_>) -> nix::Result<usize>;
 
     /// Create file node.
     async fn mknod(
         &self,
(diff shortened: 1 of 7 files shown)


Reto911:
Related issues:
- #495
- #10 
- #28

frostyplanet:
Current worker flow in this PR:
`setup_buffer_pool -> (10 buffer)
-> fuse_device_reader  -> run loop request_rx [(file, byte_buffer, size)]  -> handle_fuse_request_res()  ->spawn`

Questions:
1. fuse_device_reader() is able to read the `File`  concurrently?  
2. loop request_rs  does nothing only to pass the request to handle_fuse_request_res() , which can be put into fuse_device_reader(), thus elimate a channel passing.
3.  handle_fuse_request_res() also does nothing only to spawn(),  which can be put into fuse_device_reader()
4. I think it might be cleaner to Remove `File` from ReplyRaw,  move the reponse writing to dedicated thread, thus remove the lifetime mark

Reto911:
> Current worker flow in this PR: `setup_buffer_pool -> (some preallocated buffers) -> fuse_device_reader -> run loop request_rx [(file, byte_buffer, size)] -> handle_fuse_request_res() ->spawn`
> 
> Questions:
> 
>     1. fuse_device_reader() is able to read the `File`  concurrently?
> 
>     2. loop request_rs  does nothing only to pass the request to handle_fuse_request_res() , which can be put into fuse_device_reader(), thus elimate a channel passing.
> 
>     3. handle_fuse_request_res() also does nothing only to spawn(),  which can be put into fuse_device_reader()
> 
>     4. I think it might be cleaner to Remove `File` from ReplyRaw,  move the reponse writing to dedicated thread, thus remove the lifetime mark

1. Multiple threads cannot read a SINGLE `File` concurrently. However, in the case of ours, the two reader threads read requests from different FUSE worker `fd`s, they are cloned by `ioctl` but not via `dup`. In this case, the kernel will not lock the FUSE device on file-level, but a fine grit lock inside FUSE's kernel module, which makes the threads can read concurrently.
2. `fuse_device_reader` is not within tokio's async runtime context, therefore, we cannot spawn a task or invoke any async function here. To make the received requests can be handled within async context, we must pass the result to an async task via a channel.
3. Same as above, we cannot spawn an async task in `fuse_device_reader`.
4. I think there is no need to move the writing of response into a dedicated thread, because this writing is just a light-weight operation. Writing response in place (in tokio worker threads) can make full use of the spawned threads, which may be better to spawn dedicated threads. If we need to remove the lifetime mark from `Reply*`, I can make the `&mut File` as a parameter of `send_raw_message` method, thus the `ReplyEmpty` struct will not hold any `File` and `fd`, and `Reply*` structs will just perform as an encoder.

frostyplanet (approved):
LGTM

PR #511, opened 2024-04-01
feat: Introduce `EvictPolicy` and `Backend` for new storage

Reto911:
In this PR, we introduced `EvictPolicy` and `Backend` for new storage, designed by @xiaguan.

Besides, we added some config items for it.

### DIFF
--- a/src/new_storage/backend/backend_impl.rs
+++ b/src/new_storage/backend/backend_impl.rs
@@ -1,0 +1,126 @@
+//! The general backend implementation with `openDAL`
+
+use async_trait::async_trait;
+use opendal::raw::oio::ReadExt;
+use opendal::services::{Fs, Memory};
+use opendal::{ErrorKind, Operator};
+use tokio::io::AsyncWriteExt;
+
+use super::{Backend, StorageResult};
+
+/// The `BackendImpl` struct represents a backend storage system that implements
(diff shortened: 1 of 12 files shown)


src/new_storage/backend/backend_impl.rs
```diff
@@ -0,0 +1,103 @@
...
+    async fn store(&self, path: &str, buf: &[u8]) -> StorageResult<()> {
+        let mut writer = self.operator.writer(path).await?;
+        writer.write_all(buf).await?;
+        writer.close().await?;
+        Ok(())
+    }
+
+    #[inline]
+    async fn remove(&self, path: &str) -> StorageResult<()> {
+        self.operator.remove_all(path).await?;
```
xiaguan, line 66: I think we should change the declaration of `remove` because it actually calls `remove_all`.

Reto911: Yes, and I can add `remove_file` to perform the removing of a whole file. 

By the way, `MockIO` trait (renamed to `Storage` in my branch, and previous `Storage` is renamed to `StorageManager` for clearer naming) does not provide `remove` method for remove a file from the storage.

xiaguan (approved):

frostyplanet (approved):

PR #520, opened 2024-05-07
feat(cache node): enhance etcd client for cluster manager.

IRONICBo:
Now there are some nodes and an etcd server, the roles of these nodes are divided into master and slave, master needs to observe the list of all nodes and then update the result of hashing. slave needs to observe whether the master on the etcd is online or not, if not then it will try to compete to become the master, i.e.

In addition, all the keys registered with etcd are with lease, you need to take the initiative to renew the lease. In addition, when the client renews the lease, the lease may be disconnected, and then the client needs to re-apply for the lease and continue to renew the lease, and these states need to be marked. 

In this PR, we try to define some node roles in cache nodes, and try to enhance current etcd client to support keepalive and campaign.

Design Doc: https://datenlord.feishu.cn/docx/LoavdI51IopRuPx8Kyuc2DiRnr6

### DIFF
--- a/src/async_fuse/memfs/kv_engine/etcd_impl.rs
+++ b/src/async_fuse/memfs/kv_engine/etcd_impl.rs
@@ -1,18 +1,341 @@
-use std::collections::HashMap;
 use std::fmt::Debug;
+use std::pin::Pin;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::task::Poll;
 use std::time::Duration;
+use std::vec;
+use std::{collections::HashMap, sync::Arc};
 
+use anyhow::Context;
 use async_trait::async_trait;
+use datenlord::common::task_manager::{TaskName, TASK_MANAGER};
 use datenlord::metrics::KV_METRICS;
 use etcd_client::{
     Compare, CompareOp, DeleteOptions, GetOptions, LockOptions, PutOptions, Txn, TxnOp,
+    TxnOpResponse,
 };
+use futures::{task, Stream, StreamExt};
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info, warn};
 
 use super::{
     check_ttl, conv_u64_sec_2_i64, fmt, DeleteOption, KVEngine, KeyType, KvVersion, LockKeyType,
     MetaTxn, SetOption, ValueType,
 };
-use crate::common::error::{Context, DatenLordResult};
+use crate::common::error::DatenLordResult;
+
+/// Revoke the lease
+macro_rules! revoke_lease {
+    ($client:expr, $lease_id:expr) => {
+        match $client.lease_revoke($lease_id).await {
+            Ok(_) => {
+                info!("lease revoke success, lease_id={}", $lease_id);
+            }
+            Err(e) => {
+                warn!("failed to revoke lease, error={:?}", e);
+            }
+        }
+    };
+}
+
+/// The keepalive session struct
+#[derive(Clone)]
+pub struct Session {
+    /// close_tx is the sender to close the keep alive session
+    close_tx: mpsc::Sender<()>,
+    /// The session inner
+    inner: Arc<SessionInner>,
+}
+
+impl Debug for Session {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.inner.fmt(f)
+    }
+}
+
+impl Session {
+    /// Create a new session
+    pub async fn new(
+        client: etcd_client::Client,
+        ttl: u64,
+        lease_id: i64,
+    ) -> DatenLordResult<Arc<Self>> {
+        let closed = AtomicBool::new(false);
+        let inner = Arc::new(SessionInner {
+            closed,
+            lease_id,
+            ttl,
+            client,
+        });
+        let (close_tx, close_rx) = mpsc::channel(1);
+
+        let inner_clone = Arc::clone(&inner);
+        TASK_MANAGER
+            .spawn(TaskName::EtcdKeepAlive, |token| async move {
+                inner_clone.keepalive(close_rx, token).await;
+            })
+            .await
+            .with_context(|| "failed to spawn keep alive session task")?;
+
+        let session = Session { close_tx, inner };
+        Ok(Arc::new(session))
+    }
+
+    /// Get the lease id
+    #[must_use]
+    pub fn lease_id(&self) -> i64 {
+        self.inner.lease_id
+    }
+
+    /// Get the ttl
+    #[must_use]
+    pub fn ttl(&self) -> u64 {
+        self.inner.ttl
+    }
+
+    /// Get the `is_closed` flag
+    #[must_use]
+    pub fn is_closed(&self) -> bool {
+        self.inner.is_closed()
+    }
+
+    /// Try to close session
+    pub fn close(&self) {
+        match self.close_tx.try_send(()) {
+            Ok(o) => {
+                info!("close the keep alive session, {o:?}");
+            }
+            Err(e) => {
+                error!("failed to close the keep alive session, error={e:?}");
+            }
+        }
+        self.inner.close();
+    }
+}
+
+impl Drop for Session {
+    fn drop(&mut self) {
+        // Set the close flag
+        self.close();
+    }
+}
+
+/// The keepalive session inner struct
+pub struct SessionInner {
+    /// The closed flag
+    closed: AtomicBool,
+    /// Current lease id
+    lease_id: i64,
+    /// The ttl of the lease
+    ttl: u64,
+    /// The etcd client
+    #[cfg_attr(not(debug_assertions), skip)]
+    client: etcd_client::Client,
+}
+
+impl Debug for SessionInner {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Session")
+            .field("lease_id", &self.lease_id)
+            .field("ttl", &self.ttl)
+            .finish_non_exhaustive()
+    }
+}
+
+impl SessionInner {
+    /// Get the closed flag
+    fn is_closed(&self) -> bool {
+        self.closed.load(Ordering::Acquire)
+    }
+
+    /// Set close flag
+    fn close(&self) {
+        self.closed.store(true, Ordering::Release);
+    }
+
+    /// Keep alive the lease
+    #[allow(clippy::pattern_type_mismatch)] // Raised by `tokio::select`
+    pub async fn keepalive(
+        self: Arc<Self>,
+        mut close_rx: mpsc::Receiver<()>,
+        token: CancellationToken,
+    ) {
+        let mut interval = tokio::time::interval(Duration::from_secs(self.ttl.div_ceil(3)));
+
+        // Try to clone a client, if the keeper failed, we need to revoke the lease
+        let mut client = self.client.clone();
+        let lease_id = self.lease_id;
+        let (mut keeper, mut lease_keep_alive_stream) =
+            match client.lease_keep_alive(lease_id).await {
+                Ok((keeper, lease_keep_alive_stream)) => (keeper, lease_keep_alive_stream),
+                Err(e) => {
+                    error!("failed to keep alive lease, error={e:?}");
+
+                    // revoke the lease
+                    revoke_lease!(client, lease_id);
+                    self.close();
+                    return;
+                }
+            };
+
+        // Start to keep alive the lease
+        loop {
+            if self.is_closed() {
+                debug!("lease keep alive stream closed by is_closed flag");
+
+                // revoke the lease
+                revoke_lease!(client, lease_id);
+                return;
+            }
+
+            tokio::select! {
+                _ = interval.tick() => {
+                    // Try to send a keep alive request
+                    match keeper.keep_alive().await {
+                        Ok(o) => {
+                            debug!("keep alive lease, lease_id={lease_id}, {o:?}");
+                        }
+                        Err(e) => {
+                            error!("failed to keep alive lease, error={e:?}");
+                            break;
+                        }
+                    }
+
+                    // Try to parse
+                    match lease_keep_alive_stream.message().await {
+                        Ok(Some(lease_alive_response)) => {
+                            if lease_alive_response.ttl() == 0 {
+                                error!("lease keep alive stream closed, because ttl is 0");
+                                // revoke the lease
+                                revoke_lease!(client, lease_id);
+                                break;
+                            }
+                            continue
+                        }
+                        Ok(None) => {
+                            error!("lease keep alive stream closed, because the sender is closed");
+                            // revoke the lease
+                            revoke_lease!(client, lease_id);
+                            break;
+                        }
+                        Err(e) => {
+                            error!("failed to keep alive lease, error={e:?}");
+                            break;
+                        }
+                    }
+                }
+                _ = close_rx.recv() => {
+                    info!("close the keep alive session, lease_id={lease_id}");
+                    // revoke the lease
+                    revoke_lease!(client, lease_id);
+                    break;
+                }
+                () = token.cancelled() => {
+                    debug!("lease keep alive stream closed by token canceled");
+                    // revoke the lease
+                    revoke_lease!(client, lease_id);
+                    break;
+                }
+            }
+        }
+
+        // Make sure the lease is revoked
+        self.close();
+    }
+}
+
+/// The kvengine watch stream for etcd
+/// Try to receice message from etcd, and return key and value for each event,
+/// Wrap the etcd watch stream to support the `KVEngine` trait.
+#[derive(Debug)]
+pub struct KVEngineWatchStream {
+    /// The etcd client, used to cancel the watch stream
+    watcher: etcd_client::Watcher,
+    /// The etcd watch stream
+    watch_stream: etcd_client::WatchStream,
+}
+
+impl KVEngineWatchStream {
+    /// Create a new watch stream
+    #[must_use]
+    pub fn new(watcher: etcd_client::Watcher, watch_stream: etcd_client::WatchStream) -> Self {
+        Self {
+            watcher,
+            watch_stream,
+        }
+    }
+
+    /// Cancel the watch stream, it will take over the ownership of the watch stream
+    #[inline]
+    pub async fn cancel(mut self) -> DatenLordResult<()> {
+        match self.watcher.cancel().await {
+            Ok(o) => {
+                debug!("etcd watcher task canceled, {o:?}");
+                Ok(())
+            }
+            Err(e) => {
+                error!("failed to cancel etcd watcher, error={e:?}");
+                Err(crate::common::error::DatenLordError::EtcdClientErr {
+                    source: e,
+                    context: vec!["failed to cancel etcd watcher".to_owned()],
+                })
+            }
+        }
+    }
+}
+
+// This impl is for WatchStream from etcd_client, does testing for the watch stream.
+impl Stream for KVEngineWatchStream {
+    type Item = DatenLordResult<(String, Option<ValueType>)>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Option<Self::Item>> {
+        self.get_mut()
+            .watch_stream
+            .poll_next_unpin(cx)
+            .map(|t| match t {
+                Some(Ok(resp)) => {
+                    let mut result = None;
+                    for event in resp.events() {
+                        // Get event data
+                        if let Some(kv) = event.kv() {
+                            // Get key and value
+                            let item_key = match kv.key_str() {
+                                Ok(key) => key.to_owned(),
+                                Err(e) => {
+                                    error!("failed to get key from etcd watch event, error={e:?}");
+                                    continue;
+                                }
+                            };
+                            match event.event_type() {
+                                etcd_client::EventType::Put => {
+                                    info!("put event, key={item_key}");
+                                    let item_value = match serde_json::from_slice(kv.value()) {
+                                        Ok(value) => value,
+                                        Err(e) => {
+                                            error!("failed to deserialize value from etcd watch event, error={e:?}");
+                                            continue;
+                                        }
+                                    };
+                                    result = Some(Ok((item_key, Some(item_value))));
+                                }
+                                etcd_client::EventType::Delete => {
+                                    info!("delete event, key={item_key}");
+                                    result = Some(Ok((item_key, None)));
+                                }
+                            }
+                        }
+                    }
+                    result
+                },
+                // TODO: Cancel watch when meet an error.
+                Some(Err(e)) => Some(Err(From::from(e))),
+                None => None,
+            })
+    }
+}
 
 #[derive(Clone)]
 /// Wrap the etcd client to support the `KVEngine` trait.
@@ -62,24 +385,30 @@ impl EtcdKVEngine {
 
 #[async_trait]
 impl KVEngine for EtcdKVEngine {
+    type Session = Session;
+    type KVEngineWatchStream = KVEngineWatchStream;
+
     async fn new(end_points: Vec<String>) -> DatenLordResult<Self> {
         let client = etcd_client::Client::connect(end_points, None).await?;
         Ok(Self { client })
     }
 
-    async fn new_meta_txn(&self) -> Box<dyn MetaTxn + Send> {
+    async fn new_meta_txn(&self) -> Box<dyn MetaTxn<Session = Self::Session> + Send> {
         Box::new(EtcdTxn::new(self.client.clone()))
     }
 
-    async fn lease_grant(&self, ttl: i64) -> DatenLordResult<i64> {
+    /// Create a lease with keepalive
+    async fn create_session(&self, ttl: u64) -> DatenLordResult<Arc<Session>> {
+        let _timer = KV_METRICS.start_kv_operation_timer("create_session");
         let mut client = self.client.clone();
-        let timeout_sec = check_ttl(ttl)
-            .with_context(|| "timeout_sec should be >=1s, please fix the call".to_owned())?;
-        Ok(client
-            .lease_grant(timeout_sec, None)
+        let resp = client
+            .lease_grant(conv_u64_sec_2_i64(ttl), None)
             .await
-            .with_context(|| "failed to get lease at `MetaTxn::lock`".to_owned())?
-            .id())
+            .with_context(|| "failed to get lease at `KVEngine::create_session`".to_owned())?;
+
+        let session = Session::new(client.clone(), ttl, resp.id()).await;
+
+        return session;
     }
 
     /// Distribute lock - lock
@@ -151,8 +480,16 @@ impl KVEngine for EtcdKVEngine {
                 if option.prev_kv {
                     set_option = set_option.with_prev_key();
                 }
-                if let Some(lease) = option.lease {
-                    set_option = set_option.with_lease(lease);
+                if let Some(session) = option.session {
+                    if session.is_closed() {
+                        return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                            source: etcd_client::Error::LeaseKeepAliveError(
+                                "session is invalid".to_owned(),
+                            ),
+                            context: vec![],
+                        });
+                    }
+                    set_option = set_option.with_lease(session.lease_id());
                 }
                 Some(set_option)
             }
@@ -207,10 +544,24 @@ impl KVEngine for EtcdKVEngine {
         }
     }
 
+    /// Range get, return all key-value pairs start with prefix
     async fn range(&self, prefix: &KeyType) -> DatenLordResult<Vec<ValueType>> {
         let _timer = KV_METRICS.start_kv_operation_timer("range");
         let result = self.range_raw_key(prefix.to_string_key()).await?;
         Ok(result)
+    }
+
+    /// Watch the key, return a receiver to receive the value
+    async fn watch(&self, prefix: &KeyType) -> DatenLordResult<KVEngineWatchStream> {
+        let mut client = self.client.clone();
+        // Try to watch the key prefix
+        let opt = etcd_client::WatchOptions::new().with_prefix();
+        let (watcher, watch_stream) = client
+            .watch(prefix.to_string_key(), Some(opt.clone()))
+            .await
+            .with_context(|| "Failed to create watcher".to_owned())?;
+
+        return Ok(KVEngineWatchStream::new(watcher, watch_stream));
     }
 }
 
@@ -239,6 +590,8 @@ impl EtcdTxn {
 
 #[async_trait]
 impl MetaTxn for EtcdTxn {
+    type Session = Session;
+
     async fn get(&mut self, key_arg: &KeyType) -> DatenLordResult<Option<ValueType>> {
         let _timer = KV_METRICS.start_kv_operation_timer("get");
 
@@ -290,6 +643,122 @@ impl MetaTxn for EtcdTxn {
     fn delete(&mut self, key: &KeyType) {
         let key = key.to_string_key().into_bytes();
         self.buffer.insert(key, None);
+    }
+
+    /// Try to campaign the master with simple txn
+    /// Old(reference to etcd client v3):
+    /// Create a key with prefix, and compare key (sorted by revision) with the key created by the same session.
+    /// If current key is the smallest, then the session is the master, and campaign success.
+    ///
+    /// New(For datenlord cache scenario):
+    /// A small batch of nodes in cluster, we just need a simple txn to get the master key.
+    /// The thundering herd is not the main problem, which will cause a bunch of etcd raft logs.
+    /// Use String as val can be easily compared.
+    async fn campaign(
+        &self,
+        key: &KeyType,
+        val: String,
+        session: Arc<Session>,
+    ) -> DatenLordResult<(bool, String)> {
+        if session.is_closed() {
+            return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                source: etcd_client::Error::LeaseKeepAliveError("session is invalid".to_owned()),
+                context: vec![],
+            });
+        }
+
+        let mut client = self.client.clone();
+
+        // Try to get the key, if key is existed
+        // We need to return the data from etcd,
+        // If the key is not existed, we need to set the key.
+        let txn = Txn::new()
+            // Check the key is existed and key is not equal to current value, which means the key is ready to be campaign.
+            .when(vec![
+                Compare::create_revision(key.to_string_key(), CompareOp::NotEqual, 0),
+                Compare::value(key.to_string_key(), CompareOp::NotEqual, val.clone()),
+            ])
+            // Campaign failed
+            .and_then(vec![TxnOp::get(
+                key.to_string_key(),
+                Some(GetOptions::new().with_serializable()),
+            )])
+            // Campaign succeed
+            .or_else(vec![TxnOp::put(
+                key.to_string_key(),
+                val.clone(),
+                Some(PutOptions::new().with_lease(session.lease_id())),
+            )]);
+
+        let (campaign_status, responses) = match client.txn(txn).await {
+            Ok(resp) => {
+                // Check the txn branch is `then` or `else`
+                // If succeeded, means the campaign is failed and can not set current data to etcd
+                // If failed, means the campaign is success, and put the current data to etcd
+                (!resp.succeeded(), resp.op_responses())
+            }
+            Err(e) => {
+                error!("failed to campaign, error={e:?}");
+                return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                    source: e,
+                    context: vec!["failed to do txn operation at `MetaTxn::campaign`".to_owned()],
+                });
+            }
+        };
+
+        // Campaign success, return the current data
+        if campaign_status {
+            return Ok((true, val.clone()));
+        }
+
+        // If the campaign is failed, we need to check the response
+        if let Some(response) = responses.first() {
+            match *response {
+                TxnOpResponse::Put(_) => {
+                    // we can directly return current data
+                    return Ok((true, val.clone()));
+                }
+                TxnOpResponse::Get(ref resp) => {
+                    debug!("failed to campaign, return the existing key");
+                    if let Some(kv) = resp.kvs().first() {
+                        let item_value = match kv.value_str() {
+                            Ok(value) => value.to_owned(),
+                            Err(err) => {
+                                return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                                    source: err,
+                                    context: vec![
+                                        "failed to get key from etcd txn get event".to_owned()
+                                    ],
+                                });
+                            }
+                        };
+
+                        return Ok((false, item_value));
+                    }
+                    return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                        source: etcd_client::Error::InvalidArgs(
+                            "failed to get kvs from etcd txn get event".to_owned(),
+                        ),
+                        context: vec![],
+                    });
+                }
+                TxnOpResponse::Delete(_) | TxnOpResponse::Txn(_) => {
+                    return Err(crate::common::error::DatenLordError::EtcdClientErr {
+                        source: etcd_client::Error::InvalidArgs(
+                            "failed to campaign, the responses op is not match".to_owned(),
+                        ),
+                        context: vec![],
+                    });
+                }
+            }
+        }
+
+        return Err(crate::common::error::DatenLordError::EtcdClientErr {
+            source: etcd_client::Error::InvalidArgs(
+                "failed to campaign, the responses length is not match".to_owned(),
+            ),
+            context: vec![],
+        });
     }
 
     async fn commit(&mut self) -> DatenLordResult<bool> {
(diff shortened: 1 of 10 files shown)


src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,191 @@
...
+        self.inner.lease_id
+    }
+
+    /// Get the ttl
+    pub fn ttl(&self) -> i64 {
+        self.inner.ttl
+    }
+
+    /// Get the is_closed flag
+    pub fn is_closed(&self) -> bool {
```
xiaguan, line 87: Same name with Self::is_closed.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,191 @@
...
+                        error!("failed to keep alive lease, error={e:?}");
+
+                        // Retry to connect
+                        tokio::time::sleep(Duration::from_secs(1)).await;
+                        continue;
+                    }
+                };
+
+            // Set the is_closed flag
+            self.is_closed
```
xiaguan, line 132: Add an interface called `close()`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,191 @@
...
+                            Err(e) => {
+                                error!("failed to keep alive lease, error={e:?}");
+                                break;
+                            }
+                        }
+
+                        // Try to parse
+                        match lease_keep_alive_stream.message().await {
+                            Ok(Some(val)) => {
+                                if val.ttl() == 0 {
```
xiaguan, line 158: `if val.expired()` is a better name?
And the name of 'val` is lack of information, better renmae it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,191 @@
...
+                                    error!("lease keep alive stream closed");
+                                    self.is_closed
```
xiaguan, line 160: Same above.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +439,68 @@ impl KVEngine for EtcdKVEngine {
...
+        // Try to watch the key prefix
+        let opt = etcd_client::WatchOptions::new().with_prefix();
+        let (_watcher, mut watch_stream) = client
+            .watch(prefix.to_string_key(), Some(opt.clone()))
+            .await
+            .with_context(|| "Failed to create watcher".to_owned())?;
+
+        let self_prefix = prefix.to_string_key();
+        // Spawn a new task to handle the watch stream
+        tokio::spawn(async move {
```
xiaguan, line 467: This task should be taken over by the task manager.

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -225,8 +230,20 @@ pub trait KVEngine: Send + Sync + Debug + Sized {
     /// Lease grant
     async fn lease_grant(&self, ttl: i64) -> DatenLordResult<i64>;
 
+    /// Campaign leader
+    async fn campaign(&self, key: &KeyType, val: &ValueType, lease_id: i64) -> bool;
```
xiaguan, line 234: Why not implement it using MetaTxn?

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -225,8 +230,20 @@ pub trait KVEngine: Send + Sync + Debug + Sized {
...
+    async fn create_keep_alive_session(&self, lease_id: i64, ttl: i64) -> Arc<Self::Session>;
+
     /// Range get, return all key-value pairs start with prefix
     async fn range(&self, prefix: &KeyType) -> DatenLordResult<Vec<ValueType>>;
+
+    /// Watch the key, return a receiver to receive the value
+    async fn watch(
+        &self,
+        key: &KeyType,
+    ) -> DatenLordResult<Arc<mpsc::Receiver<(String, Option<ValueType>)>>>;
```
xiaguan, line 246: The comments for the returned tuple need to be improved.

xiaguan (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: OK, I will change it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Done.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Your are right, a `expired` can be used to check stream is simple to read.

src/async_fuse/memfs/kv_engine/mod.rs
IRONICBo: I missed this part of the implementation, I'll fix it!

src/async_fuse/memfs/kv_engine/mod.rs
IRONICBo: Okay, is it necessary to implement a struct to contain these receiver?

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: `ttl` is a library function defined in `/home/lvbo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/etcd-client-0.11.1/src/rpc/lease.rs`.

src/async_fuse/memfs/kv_engine/mod.rs
xiaguan: If watch() is not used frequently, I think it is unnecessary.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,209 @@
...
+    /// Get the ttl
+    pub fn ttl(&self) -> i64 {
+        self.inner.ttl
+    }
+
+    /// Get the is_closed flag
+    pub fn is_closed(&self) -> bool {
+        self.inner
+            .is_closed
+            .load(std::sync::atomic::Ordering::Relaxed)
```
Reto911, line 71: Should we really use `Relaxed`?

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,209 @@
...
+            .field("ttl", &self.ttl)
+            .finish()
+    }
+}
+
+impl SessionInner {
+    /// Get the is_closed flag
+    fn is_closed(&self) -> bool {
+        self.is_closed
+            .load(std::sync::atomic::Ordering::Relaxed)
```
Reto911, line 112: Same as above

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,209 @@
...
+    }
+
+    /// Set close flag
+    fn close(&self) {
+        self.is_closed
+            .store(true, std::sync::atomic::Ordering::Relaxed);
```
Reto911, line 118: Same as above

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,209 @@
...
+    }
+
+    /// Reopen close flag
+    fn reopen(&self) {
+        self.is_closed
+            .store(false, std::sync::atomic::Ordering::Relaxed);
```
Reto911, line 124: Same as above

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,19 +1,209 @@
...
+    }
+
+    /// Keep alive the lease
+    pub async fn keep_alive(&self, mut close_rx: mpsc::Receiver<()>) -> DatenLordResult<()> {
```
Reto911, line 128: It seems that an error would never be returned from this method.

IRONICBo: It's a loop function, we will try to keep it as daemon task, so we don't return an error directly.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Updated it to `AcqRel`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Updated it to `AcqRel`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Updated it to `AcqRel`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
Reto911: No, you cannot use `AcqRel` at `load` or `store`. You should use `Acquire` at `load` and use `Release` at `store`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: OK, I add it to the global task manager.

src/storage/cache_proxy/ring.rs
```diff
@@ -170,6 +170,25 @@ where
...
+    /// Create a new hash ring with current given hash builder and capacity
+    pub fn dump(&self) -> Self {
+        Self {
+            // Use the default hash builder
+            hash_builder: self.hash_builder.clone(),
+            slots: self.slots.clone(),
+            capacity: self.capacity,
+            version: self.version,
+        }
+    }
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/storage/cache_proxy/node.rs
```diff
@@ -57,46 +98,69 @@ impl Node {
...
-        let mut list = self.inner.lock().unwrap();
-        list.retain(|node| node.ip() != ip);
+    /// Get self clone
+    pub fn dump(&self) -> Self {
+        Self {
+            ip: self.ip.clone(),
+            port: self.port,
+            weight: self.weight,
+            status: self.status.clone(),
+        }
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Remove the result function.

Reto911: Yes, just return `()` insteads of a `Result`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,212 @@
...
+                            }
+                            Err(e) => {
+                                error!("failed to keep alive lease, error={e:?}");
+                                break;
+                            }
+                        }
+                    }
+                    _ = close_rx.recv() => {
+                        info!("close the keep alive session, lease_id={lease_id}");
+                        self.close();
```
IRONICBo, line 198: Revoke id from etcd.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,250 @@
...
+    }
+
+    /// Reopen close flag
+    fn reopen(&self) {
+        self.closed.store(false, Ordering::Release);
+    }
+
+    /// Keep alive the lease
+    #[allow(clippy::pattern_type_mismatch)] // Raised by `tokio::select`
+    pub async fn keep_alive(&self, mut close_rx: mpsc::Receiver<()>, token: CancellationToken) {
```
frostyplanet, line 149: 1. keep_alive ->  keepalive
2. no need to retry on lease_keepalive failed: a) the looping logic is too complex and hard to make it right. b) Once error occur lease is likely already invalided by server, retry might be meaning less.  c) Only do retrying if you can distinguish error type etcd returned, otherwise retrying might have no end condition

IRONICBo: 1. reference function is: etcd-client-0.11.1/src/client.rs
2. Thanks, I have fixed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+    /// close_tx is the sender to close the keep alive session
+    close_tx: mpsc::Sender<()>,
+    /// The session inner
+    inner: Arc<SessionInner>,
+}
+
+impl Debug for Session {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Session").finish()
```
frostyplanet, line 36: equals to ` write!(f, "Session")` ?

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -225,8 +232,24 @@ pub trait KVEngine: Send + Sync + Debug + Sized {
     /// Lease grant
     async fn lease_grant(&self, ttl: i64) -> DatenLordResult<i64>;
 
+    /// Revoke a lease
+    async fn lease_revoke(&self, lease_id: i64) -> DatenLordResult<()>;
+
+    /// Lease keep alive
+    async fn create_keep_alive_session(
```
frostyplanet, line 239: The name can be simplify to `create_session()` because session is always keepalived. It would be better to move `lease_grant()` function inside to make interface cleaner.

src/async_fuse/memfs/kv_engine/mod.rs
```diff
@@ -225,8 +232,24 @@ pub trait KVEngine: Send + Sync + Debug + Sized {
     /// Lease grant
     async fn lease_grant(&self, ttl: i64) -> DatenLordResult<i64>;
```
frostyplanet, line 233: lease_grant() can be removed because we have abstract Session.  This function is not used in the code.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+    }
+
+    /// Set close flag
+    fn close(&self) {
+        self.closed.store(true, Ordering::Release);
+    }
+
+    /// Reopen close flag
+    fn reopen(&self) {
+        self.closed.store(false, Ordering::Release);
```
frostyplanet, line 144: There's no point a invalid session can be reopen, closed should be initially false

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+    }
+
+    /// Keep alive the lease
+    #[allow(clippy::pattern_type_mismatch)] // Raised by `tokio::select`
+    pub async fn keepalive(&self, mut close_rx: mpsc::Receiver<()>, token: CancellationToken) {
+        let mut interval = tokio::time::interval(Duration::from_secs(self.ttl.div_ceil(3)));
+
+        if self.is_closed() {
```
frostyplanet, line 152: Remove `if self.is_closed`, session is only close by this functionm, or otherwise dropped (which trigger close_rx)

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+    }
+
+    /// Reopen close flag
+    fn reopen(&self) {
+        self.closed.store(false, Ordering::Release);
+    }
+
+    /// Keep alive the lease
+    #[allow(clippy::pattern_type_mismatch)] // Raised by `tokio::select`
+    pub async fn keepalive(&self, mut close_rx: mpsc::Receiver<()>, token: CancellationToken) {
```
frostyplanet, line 149: For a background task of inner struct, self can be changed to `self: Arc<Self>`

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                }
+            };
+
+        // Set the is_closed flag
+        self.reopen();
+
+        // Used to revoke client
+        let mut client = self.client.clone();
+        // Start to keep alive the lease
+        loop {
```
frostyplanet, line 176: Add a label for break to make it more clear

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                        }
+                    }
+
+                    // Try to parse
+                    match lease_keep_alive_stream.message().await {
+                        Ok(Some(lease_alive_response)) => {
+                            if lease_alive_response.ttl() == 0 {
+                                error!("lease keep alive stream closed");
+                                self.close();
+                                return;
```
frostyplanet, line 201: Replace `return` with `break`,  all breaking points should always call `self.close()`

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                            }
+                            continue
+                        }
+                        Ok(None) => {
+                            error!("lease keep alive stream closed");
```
frostyplanet, line 206: Should confirm with the etcd client code, why there're two case to indicate stream close ?  
 `Ok(None)` and `lease_alive_response.ttl == 0 `

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                _ = close_rx.recv() => {
+                    info!("close the keep alive session, lease_id={lease_id}");
+                    self.close();
+
+                    match client.lease_revoke(lease_id).await {
+                        Ok(_) => {
+                            info!("lease revoke success, lease_id={lease_id}");
+                        }
+                        Err(e) => {
+                            error!("failed to revoke lease, error={e:?}");
```
frostyplanet, line 224: Chance error to warn

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                    }
+                    return;
+                }
+                () = token.cancelled() => {
+                    match client.lease_revoke(lease_id).await {
+                        Ok(_) => {
+                            info!("lease revoke success, lease_id={lease_id}");
+                        }
+                        Err(e) => {
+                            error!("failed to revoke lease, error={e:?}");
```
frostyplanet, line 235: change error to warn

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                        }
+                    }
+                    return;
+                }
+                () = token.cancelled() => {
+                    match client.lease_revoke(lease_id).await {
+                        Ok(_) => {
+                            info!("lease revoke success, lease_id={lease_id}");
```
frostyplanet, line 232: change info to debug

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                        }
+                    }
+                }
+                _ = close_rx.recv() => {
+                    info!("close the keep alive session, lease_id={lease_id}");
+                    self.close();
+
+                    match client.lease_revoke(lease_id).await {
+                        Ok(_) => {
+                            info!("lease revoke success, lease_id={lease_id}");
```
frostyplanet, line 221: change info to debug

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+                error!("lease keep alive stream closed by is_closed flag");
+                return;
+            }
+
+            tokio::select! {
+                _ = interval.tick() => {
+                    // Try to send a keep alive request
+                    match keeper.keep_alive().await {
+                        Ok(o) => {
+                            info!("keep alive lease, lease_id={lease_id}, {o:?}");
```
frostyplanet, line 187: change info to debug

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/mod.rs
frostyplanet: I think mpsc::Recevier should not be wrapped in Arc.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I will update it.

src/async_fuse/memfs/kv_engine/mod.rs
IRONICBo: Ok, now it is removed.

src/async_fuse/memfs/kv_engine/mod.rs
IRONICBo: Ok, done.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +449,121 @@ impl KVEngine for EtcdKVEngine {
...
+                            }
+                        }
+                        break;
+                    }
+                    response = watch_stream.message() => {
+                        if let Ok(Some(watch_response)) = response {
+                            for event in watch_response.events() {
+                                // Get event data
+                                let kv = event.kv().unwrap_or_else(|| {
+                                    error!("failed to get kv from etcd watch event");
```
frostyplanet, line 509: 1. panic is dangerouse unless you are sure the case will not happend, or etcdclient behavior has been fully tested.
2. Duplicate message with error! and panic may be unnecessary, and panic can be logged in tracing
3. If you are not sure the exact behavior of etcdclient on error condition,  define a macro to log message and panic on `cfg(debug)`, and return error on `cfg[release]` for safety on production env.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +644,58 @@ impl MetaTxn for EtcdTxn {
...
+            )])
+            .or_else(vec![TxnOp::get(
+                key.to_string_key(),
+                Some(GetOptions::new().with_serializable()),
+            )]);
+
+        let res = match client.txn(txn).await {
+            Ok(resp) => {
+                // Check the txn branch is `then` or `else`
+                if resp.succeeded() {
```
frostyplanet, line 682: Should deal with resp type and value,    because put resp means campaign success, and get resp means the key exists,  at lease return the etcd key value so that does not require a second query.

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, removed.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have removed it and revoked the lease when the keepalive operation is failed.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have removed this line directly.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, so in next loop, current task will check with `is_closed()` mark.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Here are the comments from `tonic-0.9.2/src/codec/decode.rs`, `None` means that 
 the stream was closed by sender, `ttl is 0` means that the lease was outdated or revoked.
```rs
    /// Fetch the next message from this stream.
    ///
    /// # Return value
    ///
    /// - `Result::Err(val)` means a gRPC error was sent by the sender instead
    /// of a valid response message. Refer to [`Status::code`] and
    /// [`Status::message`] to examine possible error causes.
    ///
    /// - `Result::Ok(None)` means the stream was closed by the sender and no
    /// more messages will be delivered. Further attempts to call
    /// [`Streaming::message`] will result in the same return value.
    ///
    /// - `Result::Ok(Some(val))` means the sender streamed a valid response
    /// message `val`.
```

IRONICBo: Reference link is https://github.com/etcd-io/cetcd/blob/d3e3bdbcfd9609e0a5edd5315573de564ed723cb/vendor/github.com/coreos/etcd/clientv3/lease.go#L208. @frostyplanet

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Done.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I try to make it simple by `match None` operation.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I try to return both status and key in this function.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
frostyplanet: If keepalive fail, indicate networking has problem, call revoke might be blocked for a long time

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +434,117 @@ impl KVEngine for EtcdKVEngine {
...
 
+    /// Range get, return all key-value pairs start with prefix
     async fn range(&self, prefix: &KeyType) -> DatenLordResult<Vec<ValueType>> {
         let _timer = KV_METRICS.start_kv_operation_timer("range");
         let result = self.range_raw_key(prefix.to_string_key()).await?;
         Ok(result)
     }
+
+    /// Revokes a lease
+    async fn lease_revoke(&self, lease_id: i64) -> DatenLordResult<()> {
```
frostyplanet, line 445: expose revoke might not be neccessary,  since we only need to drop Session ?

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +625,92 @@ impl MetaTxn for EtcdTxn {
...
+                                    error!("failed to get key from etcd watch event, error={e:?}");
+                                    return (false, None);
+                                }
+                            };
+
+                            return (true, Some(item_key));
+                        }
+                        return (false, None);
+                    }
+                    TxnOpResponse::Delete(_) | TxnOpResponse::Txn(_) => {
```
frostyplanet, line 700: it can simpliy use `_=>` for otherwise

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +625,92 @@ impl MetaTxn for EtcdTxn {
...
+                        if let Some(kv) = resp.kvs().first() {
+                            let item_key = match kv.key_str() {
+                                Ok(key) => key.to_owned(),
+                                Err(e) => {
+                                    error!("failed to get key from etcd watch event, error={e:?}");
+                                    return (false, None);
+                                }
+                            };
+
+                            return (true, Some(item_key));
```
frostyplanet, line 696: On get resp, should return item value, not the key (since key is always predefined).

Should add some error log for unexpected condition.

the function returning type can be defined as `Result<value,  Errtype>`.

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
frostyplanet: I mean  use `break 'Label ` will be more clear because the loop is large

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,246 @@
...
+            tokio::select! {
+                _ = interval.tick() => {
+                    // Try to send a keep alive request
+                    match keeper.keep_alive().await {
+                        Ok(o) => {
+                            debug!("keep alive lease, lease_id={lease_id}, {o:?}");
+                        }
+                        Err(e) => {
+                            error!("failed to keep alive lease, error={e:?}");
+                            break;
```
frostyplanet, line 191: Remember to set close=true on all the breaking and returning branches

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +434,117 @@ impl KVEngine for EtcdKVEngine {
...
+            .with_context(|| "failed to revoke lease".to_owned())?;
+        Ok(())
+    }
+
+    /// Watch the key, return a receiver to receive the value
+    #[allow(clippy::pattern_type_mismatch)] // Raised by `tokio::select`
+    async fn watch(
+        &self,
+        prefix: &KeyType,
+    ) -> DatenLordResult<Arc<mpsc::Receiver<(String, Option<ValueType>)>>> {
```
frostyplanet, line 459: Remove the Arc on mpsc::Receiver,   since mpsc dose not allow multi receiver.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +625,92 @@ impl MetaTxn for EtcdTxn {
...
+    /// Create a key with prefix, and compare key (sorted by revision) with the key created by the same session.
+    /// If current key is the smallest, then the session is the master, and campaign success.
+    ///
+    /// New(For datenlord cache scenario):
+    /// A small batch of nodes in cluster, we just need a simple txn to get the master key.
+    /// The thundering herd is not the main problem, which will cause a bunch of etcd raft logs.
+    async fn campaign(
+        &self,
+        key: &KeyType,
+        val: &ValueType,
```
frostyplanet, line 639: valueType should be String, more suitable for comparing

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Thanks, but it can not pass the clippy if use `_ => `.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, that make sense.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have changed type to `DatenLordResult<(bool, Option<String>)>`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I call the close() function here.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, that make sense, I have removed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Maybe implement `Cmp` for `ValueType` is better?

But in current step, I have changed it to `String`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,247 @@
...
+    fn drop(&mut self) {
+        // Set the close flag
+        self.close();
+    }
+}
+
+/// The keepalive session inner struct
+pub struct SessionInner {
+    /// The closed flag
+    closed: Arc<AtomicBool>,
```
frostyplanet, line 111: closed flag does not need Arc

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,247 @@
...
+                Err(e) => {
+                    error!("failed to keep alive lease, error={e:?}");
+
+                    // revoke the lease
+                    match client.lease_revoke(lease_id).await {
+                        Ok(_) => {
+                            info!("lease revoke success, lease_id={lease_id}");
+                        }
+                        Err(e) => {
+                            error!("failed to revoke lease, error={e:?}");
```
frostyplanet, line 167: use warnning here

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,247 @@
...
+
+                    self.close();
+                    return;
+                }
+            };
+
+        // Start to keep alive the lease
+        loop {
+            if self.is_closed() {
+                error!("lease keep alive stream closed by is_closed flag");
```
frostyplanet, line 179: use debug! here. This is a normal condition.
when is_closed() == true also need to revoke.
Since there's too many duplicate code, put revoke into macro_rules.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +435,107 @@ impl KVEngine for EtcdKVEngine {
...
+
+        // Spawn a new task to handle the watch stream
+        TASK_MANAGER.spawn(TaskName::EtcdKeepAlive, |token| async move {
+            loop {
+                tokio::select! {
+                    () = token.cancelled() => {
+                        // Cancel this watch stream
+                        match watcher.cancel().await.with_context(|| "failed to cancel etcd watcher") {
+                            Ok(o) => {
+                                info!("etcd watcher task canceled, {o:?}");
```
frostyplanet, line 472: info! -> debug!

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +435,107 @@ impl KVEngine for EtcdKVEngine {
...
+                                        }
+                                    };
+                                    match event.event_type() {
+                                        etcd_client::EventType::Put => {
+                                            match tx.send((item_key, Some(item_value))).await {
+                                                Ok(o) => {
+                                                    debug!("etcd watcher put event, res={o:?}");
+                                                }
+                                                Err(e) => {
+                                                    error!("failed to send etcd watcher put event, error={e:?}");
```
frostyplanet, line 510: Ignore tx.send() error case, no need for error log.  since rx is hold by caller code

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +435,107 @@ impl KVEngine for EtcdKVEngine {
...
+
+                                            }
+                                        }
+                                        etcd_client::EventType::Delete => {
+                                            match tx.send((item_key, None)).await {
+                                                Ok(o) => {
+                                                    debug!("etcd watcher delete event, res={o:?}");
+                                                }
+                                                Err(e) => {
+                                                    error!("failed to send etcd watcher delete event, error={e:?}");
```
frostyplanet, line 521: same as above

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +616,102 @@ impl MetaTxn for EtcdTxn {
...
+    /// New(For datenlord cache scenario):
+    /// A small batch of nodes in cluster, we just need a simple txn to get the master key.
+    /// The thundering herd is not the main problem, which will cause a bunch of etcd raft logs.
+    /// Use String as val can be easily compared.
+    async fn campaign(
+        &self,
+        key: &KeyType,
+        val: String,
+        lease_id: i64,
+    ) -> DatenLordResult<(bool, Option<String>)> {
```
frostyplanet, line 633: No need to use `Option<String>` , just `String` on normal case, otherwise should be treeted as `Err()`

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +616,102 @@ impl MetaTxn for EtcdTxn {
...
+                                }
+                            };
+
+                            return Ok((false, Some(item_value)));
+                        }
+                        return Ok((false, None));
+                    }
+                    TxnOpResponse::Delete(_) | TxnOpResponse::Txn(_) => {
+                        error!("failed to campaign, the responses op is not match");
+                        return Ok((false, None));
```
frostyplanet, line 703: Should return Err()

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +616,102 @@ impl MetaTxn for EtcdTxn {
...
+                                                    .to_owned(),
+                                            ],
+                                        },
+                                    );
+                                }
+                            };
+
+                            return Ok((false, Some(item_value)));
+                        }
+                        return Ok((false, None));
```
frostyplanet, line 699: Should return Err()

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -292,6 +616,102 @@ impl MetaTxn for EtcdTxn {
...
+
+        let (campaign_status, responses) = match client.txn(txn).await {
+            Ok(resp) => {
+                // Check the txn branch is `then` or `else`
+                // If succeeded, means the campaign is failed and can not set current data to etcd
+                // If failed, means the campaign is success, and put the current data to etcd
+                (!resp.succeeded(), resp.op_responses())
+            }
+            Err(e) => {
+                error!("failed to campaign, error={e:?}");
```
frostyplanet, line 665: Here encounter etcd error,  return Err() directly, no need to fall through,

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, that's helpful, I have updated it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have fixed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have fixed it, we may aggregate `close` and `revoke` operation here.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have fixed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have fixed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have fixed it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have updated it to err.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, so we just need to treat `campaign failed` as Err.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -529,4 +1045,110 @@ mod test {
...
+        let key = KeyType::String("test_campaign".to_owned());
+        let val = "test_campaign_val".to_owned();
+        let val2 = "test_campaign_val2".to_owned();
+        let lease_id = client.create_lease(5).await.unwrap();
+
+        // Campaign the key
+        let (campaign_status, campaign_val) = txn.campaign(&key, val.clone(), lease_id).await.unwrap();
+        assert_eq!(campaign_status, true);
+        assert_eq!(campaign_val, val);
+
```
frostyplanet, line 1143: 2nd time: Should use a different session/lease_id. Expect campain to overwrite the value,  and then drop the original session.
Maybe test query the key and confirm new lease_id.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -529,4 +1045,110 @@ mod test {
...
+        // Recampaign the key
+        let (campaign_status, campaign_val) = txn.campaign(&key, val.clone(), lease_id).await.unwrap();
+        assert_eq!(campaign_status, true);
+        assert_eq!(campaign_val, val);
+
```
frostyplanet, line 1148: 3rd time:  Should use a different session/lease_id, and make sure 2nd session is not dropped.

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,396 @@
...
+                    source: e,
+                    context: vec!["failed to cancel etcd watcher".to_owned()],
+                })
+            }
+        }
+    }
+
+    /// Get the message from the watch stream
+    #[inline]
+    pub async fn message(&mut self) -> DatenLordResult<Option<(String, Option<ValueType>)>> {
```
frostyplanet, line 293: Most of the code is duplicated with the stream impl.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -529,4 +1045,110 @@ mod test {
...
     }
+
+    #[tokio::test]
+    async fn test_watch() {
+        let client = EtcdKVEngine::new_for_local_test(vec![ETCD_ADDRESS.to_owned()])
+            .await
+            .unwrap();
+        // Create a new key
+        let client_clone = client.clone();
+        tokio::spawn(async move {
```
frostyplanet, line 1056: Should test the stream mode with a series of key create/update/delete

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -1,18 +1,396 @@
...
+    pub fn new(watcher: etcd_client::Watcher, watch_stream: etcd_client::WatchStream) -> Self {
+        Self {
+            watcher,
+            watch_stream,
+        }
+    }
+
+    /// Cancel the watch stream
+    #[inline]
+    pub async fn cancel(&mut self) -> DatenLordResult<()> {
```
frostyplanet, line 275: Better change `cancel`  to `async fn close(self)`, which will consume the obj ownership

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Thanks, I've updated with new `lease_id` before each `campaign`.

BTW, I've add a `get_lease_id` for test to get lease_id from `get`.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I've added it.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I've removed it, and fixed the issue "when met a `Delete` event, this implement will skip the delete error."

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I put it into a for range to test a series of operations.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: `watcher.cancel()` from `etcd_client` need a mutable value.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo:
```bash
running 1 test
2024-06-20T05:28:01.272657Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:02.275220Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:03.277604Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: delete event, key=TEST_test_watch
2024-06-20T05:28:04.280747Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:05.283861Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:06.286420Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: delete event, key=TEST_test_watch
2024-06-20T05:28:07.290238Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:08.292962Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:09.294694Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: delete event, key=TEST_test_watch
2024-06-20T05:28:10.297038Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:11.299703Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:12.302634Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: delete event, key=TEST_test_watch
2024-06-20T05:28:13.305388Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:14.307972Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: put event, key=TEST_test_watch
2024-06-20T05:28:15.309519Z  INFO datenlord::async_fuse::memfs::kv_engine::etcd_impl: delete event, key=TEST_test_watch
test async_fuse::memfs::kv_engine::etcd_impl::test::test_watch ... ok
```

src/async_fuse/memfs/kv_engine/etcd_impl.rs
frostyplanet: Should only use session, lease is a low level api

src/async_fuse/memfs/kv_engine/etcd_impl.rs
```diff
@@ -207,11 +531,37 @@ impl KVEngine for EtcdKVEngine {
...
+    /// Create a lease
+    async fn create_lease(&self, ttl: i64) -> DatenLordResult<i64> {
+        let _timer = KV_METRICS.start_kv_operation_timer("lease");
+        let ttl = check_ttl(ttl).with_context(|| "failed to check ttl".to_owned())?;
+        let mut client = self.client.clone();
+        let resp = client
+            .lease_grant(ttl, None)
+            .await
+            .with_context(|| "failed to get lease at `MetaTxn::create_lease`".to_owned())?;
+        Ok(resp.id())
```
frostyplanet, line 550: Should be `create_session`. We should not expose lease to user,  return a Session.

frostyplanet (changes requested):

src/async_fuse/memfs/kv_engine/etcd_impl.rs
frostyplanet: Use `async fn close(mut self)` then, you should consume the ownership

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have remove this function and use create_session instead.

src/async_fuse/memfs/kv_engine/etcd_impl.rs
IRONICBo: Ok, I have updated it.

frostyplanet (approved):

PR #523, opened 2024-05-31
feat(cache node): support file async rpc.

IRONICBo:
# Background
In response to the need for efficient file transfers in distributed environments, this system aims to achieve fast transfer of file blocks and file ranges through asynchronous RPC (ARPC) methods. The system will focus on the transfer of file blocks (block) or file ranges (range) to optimize data flow processing and ensure data consistency and efficient transfer. It is mainly used for IO-intensive data streams and is particularly suitable for block and file storage IO operations. The design goal is to minimize CPU usage and maximize throughput.

# Problems
1. client connection is high, client will usually request to download bulk blocks from cache node, different blocks may be mapped to different cache node nodes, so a client may need to maintain many connections. 2. possible continuous block transfer, especially suitable for block storage and file storage IO operations.
2. possible continuous block transfer, in the current load balancing algorithm, the number of blocks and cache node number of different orders of magnitude, it is likely that there are more than one block, or even a continuous block mapped to the same cache node node, frequent creation of the same peer-to-peer connection may be unnecessary overhead.
3. The server side may need to handle many client requests at the same time, and each request is mainly an io task, supplemented by protocol parsing (simple check). Therefore, we need to simplify the server-side implementation, and at the same time, we need to manage these io tasks well.

# Design
In server side, we need to implement `handler` for server conn, and put a server job to workerpool(control the limitation).

In client side, we use `packets keeper` to keep packets and manager connection is timeout for current request. Also, we use `seq` to mark each request in one stream.

> Ref: https://datenlord.feishu.cn/docx/G7LTdqL3koUo8wxJ1aucHeELnNd

### DIFF
--- a/src/storage/cache/rpc/client.rs
+++ b/src/storage/cache/rpc/client.rs
@@ -1,0 +1,804 @@
+use std::{
+    cell::UnsafeCell,
+    fmt::Debug,
+    future::IntoFuture,
+    pin::Pin,
+    sync::{atomic::AtomicU64, Arc},
+    task::{Context, Poll},
+};
+
+use bytes::BytesMut;
+use futures::{pin_mut, Future};
+use tokio::{
+    io::{AsyncRead, AsyncWriteExt, ReadBuf},
+    net::TcpStream,
+    sync::oneshot,
+    time::{Instant, Interval},
+};
+use tracing::debug;
+
+use crate::write_all_timeout;
+
+use super::{
+    common::{self, ClientTimeoutOptions},
+    error::RpcError,
+    message::{ReqType, RespType},
+    packet::{Decode, Encode, Packet, PacketsKeeper, ReqHeader, RespHeader, REQ_HEADER_SIZE},
+    utils::u64_to_usize,
+};
+
+/// TODO: combine `RpcClientConnectionInner` and `RpcClientConnection`
+struct RpcClientConnectionInner<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// The TCP stream for the connection.
+    stream: UnsafeCell<TcpStream>,
+    /// Options for the timeout of the connection
+    timeout_options: ClientTimeoutOptions,
+    /// Stream auto increment sequence number, used to mark the request and response
+    seq: AtomicU64,
+    /// The instant of the connection, used to calculate the keep alive timeout
+    clock_instant: Instant,
+    /// Received keep alive timestamp, will mark the valid keep alive response.
+    /// If the keep alive response is not received in right mestamp, the connection will be closed,
+    /// If we receive other response, we will update the received_keepalive_timestamp too, just treat it as a keep alive response.
+    received_keepalive_timestamp: AtomicU64,
+    /// Send packet task
+    packets_keeper: PacketsKeeper<P>,
+    /// Response buffer, try to reuse same buffer to reduce memory allocation
+    /// In case of the response is too large, we need to consider the buffer size
+    /// Init size is 4MB
+    /// Besides, we want to reduce memory copy, and read/write the response to the same data buffer
+    resp_buf: UnsafeCell<BytesMut>,
+    /// Request encode buffer
+    req_buf: UnsafeCell<BytesMut>,
+    /// The Client ID for the connection
+    client_id: u64,
+    /// The receiver for the closed signal.
+    is_closed_receiver: UnsafeCell<oneshot::Receiver<()>>,
+}
+
+// TODO: Add markable id for this client
+impl<P> Debug for RpcClientConnectionInner<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RpcClientConnectionInner")
+            .field("client id", &self.client_id)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<P> RpcClientConnectionInner<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// Create a new connection.
+    pub fn new(
+        stream: TcpStream,
+        timeout_options: ClientTimeoutOptions,
+        client_id: u64,
+        is_closed_receiver: oneshot::Receiver<()>,
+    ) -> Self {
+        let task_timeout = timeout_options.task_timeout.as_secs();
+        let is_closed_receiver_future = is_closed_receiver.into_future();
+        Self {
+            stream: UnsafeCell::new(stream),
+            timeout_options,
+            seq: AtomicU64::new(0),
+            clock_instant: Instant::now(),
+            received_keepalive_timestamp: AtomicU64::new(0),
+            packets_keeper: PacketsKeeper::new(task_timeout),
+            resp_buf: UnsafeCell::new(BytesMut::with_capacity(
+                common::DEFAULT_TCP_RESPONSE_BUFFER_SIZE,
+            )),
+            req_buf: UnsafeCell::new(BytesMut::with_capacity(
+                common::DEFAULT_TCP_REQUEST_BUFFER_SIZE,
+            )),
+            client_id,
+            is_closed_receiver: UnsafeCell::new(is_closed_receiver_future),
+        }
+    }
+
+    /// Recv request header from the stream
+    async fn recv_header(&self) -> Result<RespHeader, RpcError> {
+        // Try to read to buffer
+        match self.recv_len(REQ_HEADER_SIZE).await {
+            Ok(()) => {}
+            Err(err) => {
+                debug!("Failed to receive request header: {:?}", err);
+                return Err(err);
+            }
+        }
+
+        let buffer: &mut BytesMut = self.get_resp_buf_mut();
+        let req_header = RespHeader::decode(buffer)?;
+        debug!("Received response header: {:?}", req_header);
+
+        Ok(req_header)
+    }
+
+    /// Receive response body from the server.
+    async fn recv_len(&self, len: u64) -> Result<(), RpcError> {
+        let req_buffer: &mut BytesMut = self.get_resp_buf_mut();
+        req_buffer.resize(u64_to_usize(len), 0);
+
+        let recv_len_future = RecvLenFuture::new(self, len, req_buffer);
+
+        // Block here until the client is dropped or stream error.
+        match recv_len_future.await {
+            Ok(size) => {
+                debug!("Received response body size: {:?}", size);
+                Ok(())
+            }
+            Err(err) => {
+                debug!("Failed to receive response header: {:?}", err);
+                Err(RpcError::InternalError(err.to_string()))
+            }
+        }
+    }
+
+    /// Send a request to the server.
+    /// We need to make sure
+    /// Current send packet need to be in one task, so if we need to send ping and packet
+    /// we need to make sure only one operation is running, like select.
+    async fn send_data(&self, req_header: &dyn Encode, packet: Option<P>) -> Result<(), RpcError> {
+        let buf = self.get_req_buf_mut();
+        buf.clear();
+        // encode just need to append to buffer, do not clear buffer
+        req_header.encode(buf);
+        // Append the body to the buffer
+        if let Some(body) = packet {
+            // encode just need to append to buffer, do not clear buffer
+            body.encode(buf);
+        }
+        debug!("{:?} Sent data with length: {:?}", self, buf.len());
+        let writer = self.get_stream_mut();
+        match write_all_timeout!(writer, buf, self.timeout_options.write_timeout).await {
+            Ok(()) => Ok(()),
+            Err(err) => {
+                debug!("{:?} Failed to send data: {:?}", self, err);
+                Err(RpcError::InternalError(err.to_string()))
+            }
+        }
+    }
+
+    /// Get the next sequence number.
+    fn next_seq(&self) -> u64 {
+        self.seq.fetch_add(1, std::sync::atomic::Ordering::AcqRel)
+    }
+
+    /// Send keep alive message to the server
+    async fn ping(&self) -> Result<(), RpcError> {
+        // Send keep alive message
+        let current_seq = self.next_seq();
+        let current_timestamp = self.clock_instant.elapsed().as_secs();
+        // Check keepalive is valid
+        let received_keepalive_timestamp = self
+            .received_keepalive_timestamp
+            .load(std::sync::atomic::Ordering::Acquire);
+        if current_timestamp - received_keepalive_timestamp
+            > self.timeout_options.keep_alive_timeout.as_secs()
+        {
+            debug!("{:?} keep alive timeout, close the connection", self);
+            return Err(RpcError::InternalError("Keep alive timeout".to_owned()));
+        }
+
+        // Set to packet task
+        let keep_alive_msg = ReqHeader {
+            seq: current_seq,
+            op: ReqType::KeepAliveRequest.into(),
+            len: 0,
+        };
+
+        if let Ok(()) = self.send_data(&keep_alive_msg, None).await {
+            debug!("{:?} Success to sent keep alive message", self);
+            Ok(())
+        } else {
+            debug!("{:?} Failed to send keep alive message", self);
+            Err(RpcError::InternalError(
+                "Failed to send keep alive message".to_owned(),
+            ))
+        }
+    }
+
+    /// Send packet by the client.
+    async fn send_packet(&self, mut req_packet: P) -> Result<(), RpcError> {
+        let current_seq = self.next_seq();
+        req_packet.set_seq(current_seq);
+        debug!("{:?} Try to send request: {:?}", self, current_seq);
+
+        // Send may be used in different threads, so we need to create local buffer to store the request data
+        let req_len = req_packet.get_req_len(); // Try to get request buffer
+        let req_header = ReqHeader {
+            seq: req_packet.seq(),
+            op: req_packet.op(),
+            len: req_len,
+        };
+
+        // concate req_header and req_buffer
+        let req_seq = req_packet.seq();
+        if let Ok(()) = self.send_data(&req_header, Some(req_packet)).await {
+            debug!("{:?} Sent request success: {:?}", self, req_seq);
+            Ok(())
+        } else {
+            debug!("{:?} Failed to send request: {:?}", self, req_seq);
+            Err(RpcError::InternalError("Failed to send request".to_owned()))
+        }
+    }
+
+    /// Receive loop for the client.
+    async fn recv_loop(&self) {
+        // Check and clean keeper timeout unused task
+        // The timeout data will be cleaned by the get function
+        // We don't need to clean the timeout data radically
+        let mut tickers = Box::pin(tokio::time::interval(
+            self.timeout_options.task_timeout.div_f32(2.0),
+        ));
+        loop {
+            // Clean timeout tasks, will block here
+            let recv_header_f = self.recv_header();
+            pin_mut!(recv_header_f);
+            let selector = ReceiveHeaderFuture::new(self, &mut tickers, &mut recv_header_f);
+            match selector.await {
+                Ok(header) => {
+                    // Try to receive the response from the server
+                    let header_seq = header.seq;
+                    debug!("{:?} Received keep alive response or other response.", self);
+                    let current_timestamp = self.clock_instant.elapsed().as_secs();
+                    self.received_keepalive_timestamp
+                        .store(current_timestamp, std::sync::atomic::Ordering::Release);
+                    // Update the received keep alive seq
+                    if let Ok(resp_type) = RespType::try_from(header.op) {
+                        if let RespType::FileBlockResponse = resp_type {
+                            debug!("{:?} Received response header: {:?}", self, header);
+                            // Try to read to buffer
+                            match self.recv_len(header.len).await {
+                                Ok(()) => {}
+                                Err(err) => {
+                                    debug!(
+                                        "{:?} Failed to receive request header: {:?}",
+                                        self, err
+                                    );
+                                    break;
+                                }
+                            }
+
+                            // Take the packet task and recv the response
+                            let resp_buffer: &mut BytesMut = self.get_resp_buf_mut();
+                            match self.packets_keeper.take_task(header_seq, resp_buffer).await {
+                                Ok(()) => {
+                                    debug!("{:?} Received response: {:?}", self, header_seq);
+                                }
+                                Err(err) => {
+                                    debug!("{:?} Failed to update task: {:?}", self, err);
+                                }
+                            }
+                        }
+                    } else {
+                        debug!("{:?} Invalid response type: {:?}", self, header.op);
+                        break;
+                    }
+                }
+                Err(err) => {
+                    debug!("{:?} Failed to receive request header: {:?}", self, err);
+                    break;
+                }
+            }
+        }
+
+        // Purge all the tasks
+        self.packets_keeper.purge_outdated_tasks().await;
+    }
+
+    /// Get stream with mutable reference
+    #[allow(clippy::mut_from_ref)]
+    fn get_stream_mut(&self) -> &mut TcpStream {
+        // Current implementation is safe because the stream is only accessed by one thread
+        unsafe { &mut *self.stream.get() }
+    }
+
+    /// Get resp buffer with mutable reference
+    #[allow(clippy::mut_from_ref)]
+    fn get_resp_buf_mut(&self) -> &mut BytesMut {
+        unsafe { &mut *self.resp_buf.get() }
+    }
+
+    /// Get req buffer with mutable reference
+    #[allow(clippy::mut_from_ref)]
+    fn get_req_buf_mut(&self) -> &mut BytesMut {
+        unsafe { &mut *self.req_buf.get() }
+    }
+
+    /// Get the closed receiver for the connection with mutable reference.
+    #[allow(clippy::mut_from_ref)]
+    fn get_is_closed_receiver_mut(&self) -> &mut oneshot::Receiver<()> {
+        unsafe { &mut *self.is_closed_receiver.get() }
+    }
+}
+
+unsafe impl<P> Send for RpcClientConnectionInner<P> where P: Packet + Clone + Send + Sync + 'static {}
+
+unsafe impl<P> Sync for RpcClientConnectionInner<P> where P: Packet + Clone + Send + Sync + 'static {}
+
+/// The RPC client definition.
+#[derive(Debug)]
+pub struct RpcClient<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// The timeout options for the client.
+    /// TODO:
+    #[allow(dead_code)]
+    timeout_options: ClientTimeoutOptions,
+    /// Inner connection for the client.
+    inner_connection: Arc<RpcClientConnectionInner<P>>,
+    /// Client ID for the connection
+    /// TODO:
+    #[allow(dead_code)]
+    client_id: AtomicU64,
+    /// Is closed flag sender, this struct is associated with the inner_connection,
+    /// if this client is dropped, the inner_connection receiver will receive the closed signal.
+    #[allow(dead_code)]
+    is_closed_sender: oneshot::Sender<()>,
+}
+
+impl<P> RpcClient<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// Create a new RPC client.
+    ///
+    /// We don't manage the stream is dead or clean
+    /// The client will be closed if the keep alive message is not received in 100 times
+    /// When the stream is broken, the client will be closed, and we need to recreate a new client
+    pub fn new(stream: TcpStream, timeout_options: ClientTimeoutOptions) -> Self {
+        let client_id = AtomicU64::new(0);
+        // let (is_closed_sender, is_closed_receiver) = flume::unbounded::<()>();
+        let (is_closed_sender, is_closed_receiver) = oneshot::channel::<()>();
+        let inner_connection = RpcClientConnectionInner::new(
+            stream,
+            timeout_options.clone(),
+            client_id.load(std::sync::atomic::Ordering::Acquire),
+            is_closed_receiver,
+        );
+
+        Self {
+            timeout_options,
+            inner_connection: Arc::new(inner_connection),
+            client_id,
+            is_closed_sender,
+        }
+    }
+
+    /// Start client receiver
+    pub fn start_recv(&self) {
+        // Create a receiver loop
+        let inner_connection_clone = Arc::clone(&self.inner_connection);
+        tokio::spawn(async move {
+            inner_connection_clone.recv_loop().await;
+        });
+    }
+
+    /// Send a request to the server.
+    /// Try to send data to channel, if the channel is full, return an error.
+    /// Contains the rezquest header and body.
+    /// WARN: this function does not support concurrent call
+    pub async fn send_request(&self, req: P) -> Result<(), RpcError> {
+        self.inner_connection
+            .send_packet(req)
+            .await
+            .map_err(|err| RpcError::InternalError(err.to_string()))
+    }
+
+    /// Manually send ping by the send request
+    /// The inner can not start two loop, because the stream is unsafe
+    /// we need to start the loop in the client or higher level manually
+    /// WARN: this function does not support concurrent call
+    pub async fn ping(&self) -> Result<(), RpcError> {
+        self.inner_connection.ping().await
+    }
+}
+
+/// `RecvLenFuture` TCP Stream reader future for receive data
+/// We need to keep block the stream until the data is received
+/// or the stream is closed or client is dropped
+/// The future will be used to receive the response from the server
+struct RecvLenFuture<'a, P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// The inner connection for the client.
+    client: &'a RpcClientConnectionInner<P>,
+    /// The length of the data to receive.
+    len: u64,
+    /// The buffer to store the received data.
+    req_buffer: &'a mut BytesMut,
+}
+
+impl<'a, P> RecvLenFuture<'a, P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// Create a new receive length future.
+    fn new(
+        client: &'a RpcClientConnectionInner<P>,
+        len: u64,
+        req_buffer: &'a mut BytesMut,
+    ) -> Self {
+        Self {
+            client,
+            len,
+            req_buffer,
+        }
+    }
+}
+
+impl<P> Future for RecvLenFuture<'_, P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    type Output = Result<(), RpcError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+
+        // Resize the buffer to accommodate the incoming data
+        this.req_buffer.resize(u64_to_usize(this.len), 0);
+
+        let mut reader = this.client.get_stream_mut();
+        let mut read_buf = ReadBuf::new(this.req_buffer);
+
+        let buf_poll_status = match Pin::new(&mut reader).poll_read(cx, &mut read_buf) {
+            Poll::Ready(Ok(size)) => {
+                debug!("Received response body size: {:?}", size);
+                Poll::Ready(Ok(()))
+            }
+            Poll::Ready(Err(err)) => {
+                debug!("Failed to receive response header: {:?}", err);
+                Poll::Ready(Err(RpcError::InternalError(err.to_string())))
+            }
+            Poll::Pending => Poll::Pending,
+        };
+
+        // Recv closed data from the client channel
+        let is_closed_receiver_future = this.client.get_is_closed_receiver_mut();
+        match Pin::new(is_closed_receiver_future).poll(cx) {
+            Poll::Ready(Ok(())) => {
+                debug!("Client is closed, stop the receive loop");
+                Poll::Ready(Err(RpcError::InternalError("Client is closed".to_owned())))
+            }
+            Poll::Ready(Err(err)) => {
+                debug!("Failed to receive closed signal: {:?}", err);
+                Poll::Ready(Err(RpcError::InternalError(err.to_string())))
+            }
+            Poll::Pending => buf_poll_status,
+        }
+    }
+}
+
+/// The receive stream for the client.
+/// The stream will be used to receive the response from the server,
+/// and clean outdated tasks when the task is timeout.
+struct ReceiveHeaderFuture<'a, P, F>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+    F: Future<Output = Result<RespHeader, RpcError>> + Unpin,
+{
+    /// The inner connection for the client.
+    client_inner: &'a RpcClientConnectionInner<P>,
+    /// The interval for period task.
+    interval: &'a mut Pin<Box<Interval>>,
+    /// The recv_future in the client.
+    recv_future: Pin<&'a mut F>,
+}
+
+impl<'a, P, F> ReceiveHeaderFuture<'a, P, F>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+    F: Future<Output = Result<RespHeader, RpcError>> + Unpin,
+{
+    /// Create a new receive header stream.
+    fn new(
+        client_inner: &'a RpcClientConnectionInner<P>,
+        interval: &'a mut Pin<Box<Interval>>,
+        recv_future: &'a mut F,
+    ) -> Self {
+        Self {
+            client_inner,
+            interval,
+            recv_future: Pin::new(recv_future),
+        }
+    }
+}
+
+impl<'a, P, F> Future for ReceiveHeaderFuture<'a, P, F>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+    F: Future<Output = Result<RespHeader, RpcError>> + Unpin,
+{
+    type Output = Result<RespHeader, RpcError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let self_mut: &mut ReceiveHeaderFuture<'a, P, F> = self.get_mut();
+        let client_inner = self_mut.client_inner;
+
+        // clean timeout tasks
+        let mut tick_ready = false;
+        while self_mut.interval.as_mut().poll_tick(cx).is_ready() {
+            // consume all ticks in once call
+            tick_ready = true;
+        }
+        while tick_ready {
+            let clean_future = client_inner.packets_keeper.clean_timeout_tasks();
+            pin_mut!(clean_future);
+            match clean_future.poll(cx) {
+                Poll::Ready(()) => tick_ready = false,
+                Poll::Pending => return Poll::Pending,
+            }
+        }
+
+        // recv header data
+        self_mut.recv_future.as_mut().poll(cx)
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+#[allow(clippy::indexing_slicing)]
+mod tests {
+    use async_trait::async_trait;
+    use bytes::BufMut;
+    use tokio::net::TcpListener;
+
+    use crate::{async_fuse::util::usize_to_u64, storage::cache::rpc::utils::get_u64_from_buf};
+
+    use super::*;
+    use std::{mem, time::Duration};
+
+    /// Request struct for test
+    #[derive(Debug, Default, Clone)]
+    pub struct TestRequest {
+        pub id: u64,
+    }
+
+    impl Encode for TestRequest {
+        fn encode(&self, buf: &mut BytesMut) {
+            buf.put_u64(self.id.to_le());
+        }
+    }
+
+    impl Decode for TestRequest {
+        fn decode(buf: &[u8]) -> Result<Self, RpcError> {
+            if buf.len() < 8 {
+                return Err(RpcError::InternalError("Insufficient bytes".to_owned()));
+            }
+            let id = get_u64_from_buf(buf, 0)?;
+            Ok(TestRequest { id })
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    pub struct TestPacket {
+        pub seq: u64,
+        pub op: u8,
+        pub request: TestRequest,
+        pub timestamp: u64,
+    }
+
+    impl Encode for TestPacket {
+        fn encode(&self, buf: &mut BytesMut) {
+            self.request.encode(buf);
+        }
+    }
+
+    #[async_trait]
+    impl Packet for TestPacket {
+        fn seq(&self) -> u64 {
+            self.seq
+        }
+
+        fn set_seq(&mut self, seq: u64) {
+            self.seq = seq;
+        }
+
+        fn op(&self) -> u8 {
+            self.op
+        }
+
+        fn set_op(&mut self, op: u8) {
+            self.op = op;
+        }
+
+        fn set_timestamp(&mut self, timestamp: u64) {
+            self.timestamp = timestamp;
+        }
+
+        fn get_timestamp(&self) -> u64 {
+            self.timestamp
+        }
+
+        fn set_resp_data(&mut self, _data: &[u8]) -> Result<(), RpcError> {
+            Ok(())
+        }
+
+        async fn set_result(self, _status: Result<(), RpcError>) {}
+
+        fn get_req_len(&self) -> u64 {
+            usize_to_u64(mem::size_of_val(&self.request))
+        }
+    }
+
+    // Helper function to setup a mock server
+    async fn setup_mock_server(addr: String, response: Vec<u8>) -> String {
+        let listener = TcpListener::bind(addr).await.unwrap();
+        let addr = listener.local_addr().unwrap().to_string();
+        tokio::spawn(async move {
+            if let Ok((mut socket, _)) = listener.accept().await {
+                socket.write_all(&response).await.unwrap(); // Send a predefined response
+                socket.flush().await.unwrap();
+            }
+        });
+        addr
+    }
+
+    #[tokio::test]
+    async fn test_new_connection() {
+        // setup();
+        let local_addr = setup_mock_server("127.0.0.1:50050".to_owned(), vec![]).await;
+        let stream = TcpStream::connect(local_addr).await.unwrap();
+        let timeout_options = ClientTimeoutOptions {
+            read_timeout: Duration::from_secs(5),
+            write_timeout: Duration::from_secs(5),
+            task_timeout: Duration::from_secs(60),
+            keep_alive_timeout: Duration::from_secs(30),
+        };
+        let client_id = 123;
+        let (sender, receiver) = oneshot::channel::<()>();
+        let connection = RpcClientConnectionInner::<TestPacket>::new(
+            stream,
+            timeout_options,
+            client_id,
+            receiver,
+        );
+        assert_eq!(connection.client_id, client_id);
+
+        sender.send(()).unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_recv_header() {
+        // setup();
+        let mut buf = BytesMut::new();
+        RespHeader {
+            seq: 1,
+            op: RespType::KeepAliveResponse.into(),
+            len: 0,
+        }
+        .encode(&mut buf);
+        let addr = setup_mock_server("127.0.0.1:50052".to_owned(), buf.to_vec()).await;
+        let stream = TcpStream::connect(addr).await.unwrap();
+        let timeout_options = ClientTimeoutOptions {
+            read_timeout: Duration::from_secs(5),
+            write_timeout: Duration::from_secs(5),
+            task_timeout: Duration::from_secs(60),
+            keep_alive_timeout: Duration::from_secs(30),
+        };
+        let (sender, receiver) = oneshot::channel::<()>();
+        let connection =
+            RpcClientConnectionInner::<TestPacket>::new(stream, timeout_options, 123, receiver);
+        let header = connection.recv_header().await.unwrap();
+        assert_eq!(header.seq, 1);
+        let resp_op: u8 = RespType::KeepAliveResponse.into();
+        assert_eq!(header.op, resp_op);
+        sender.send(()).unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_recv_len() {
+        // setup();
+        // Create a 10 len vector
+        let response = vec![0; 10];
+        let addr = setup_mock_server("127.0.0.1:50053".to_owned(), response).await;
+        let stream = TcpStream::connect(addr).await.unwrap();
+        let timeout_options = ClientTimeoutOptions {
+            read_timeout: Duration::from_secs(5),
+            write_timeout: Duration::from_secs(5),
+            task_timeout: Duration::from_secs(60),
+            keep_alive_timeout: Duration::from_secs(30),
+        };
+        let (sender, receiver) = oneshot::channel::<()>();
+        let connection =
+            RpcClientConnectionInner::<TestPacket>::new(stream, timeout_options, 123, receiver);
+        connection.recv_len(10).await.unwrap();
+
+        sender.send(()).unwrap();
+    }
+
+    struct TestData;
+    impl Encode for TestData {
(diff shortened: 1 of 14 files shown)


frostyplanet:
PacketsKeeper  concurrently accessed by send & recv_loop,  should not use Unsafe, use mutex instead

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// Get stream with mutable reference
+    #[inline(always)]
+    fn get_stream_mut(&self) -> &mut net::TcpStream {
+        // Current implementation is safe because the stream is only accessed by one thread
+        unsafe { std::mem::transmute(self.stream.get()) }
+    }
+
+    /// Get stream with immutable reference
+    #[inline(always)]
+    fn get_stream(&self) -> &net::TcpStream {
```
frostyplanet, line 264: remove unused function

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub fn next_seq(&self) -> u64 {
+        self.seq.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+    }
+
+    pub async fn ping(&self) -> Result<(), RpcError<String>> {
+        // Send keep alive message
+        let current_seq = self.next_seq();
+        // Check keepalive is valid
+        let received_keepalive_seq = self.received_keepalive_seq.load(std::sync::atomic::Ordering::Relaxed);
+        if current_seq - received_keepalive_seq > 100 {
```
frostyplanet, line 122: Do not use seq here due to highly dependency to the upper level usage of ping(), use timestamp instead. Add an option for the timeout threshold.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub async fn recv_len(&self, len: u64) -> Result<(), RpcError<String>> {
+        let mut req_buffer: &mut BytesMut = unsafe {
+            transmute(self.resp_buf.get())
+        };
+        req_buffer.resize(len as usize, 0);
+        let reader = self.get_stream_mut();
+        match read_exact_timeout!(reader, &mut req_buffer, self.timeout_options.read_timeout).await
+        {
+            Ok(size) => {
+                debug!("Received response body len: {:?}", size);
```
frostyplanet, line 87: All log related to client should include information about "self"  (eg: destination), there're multiple rpc client instances exist in one program.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,458 @@
...
+            }.encode();
+
+            // concate req_header and req_buffer
+            req_header.extend_from_slice(&req_buffer);
+
+            if let Ok(_) = self.send_data(&req_header).await {
+                debug!("Sent request success: {:?}", req_packet.seq());
+                // We have set a copy to keeper and manage the status for the packets keeper
+                // Set to packet task with clone
+                self.get_packets_keeper_mut().add_task(req_packet.clone()).await;
```
frostyplanet, line 164: add_task() does not need to be async fn.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        // Remove the timeout packets
+        for seq in timeout_packets {
+            self.packets.remove(&seq);
+            self.timestamp.remove(&seq);
+        }
+    }
+
+    /// Get a task from the packets and remove it
+    pub async fn consume_task(&mut self, seq: u64) -> Option<P> {
```
frostyplanet, line 245: No need for async fn

frostyplanet (changes requested):

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,458 @@
...
+            tokio::select! {
+                _ = tickers.tick() => {
+                    let packets_keeper: &mut PacketsKeeper<P> = self.get_packets_keeper_mut();
+                    packets_keeper.clean_timeout_tasks().await;
+                }
+            }
+
+            // Try to receive the response from the server
+            debug!("Waiting for response...");
+            let resp_header = self.recv_header().await;
```
frostyplanet, line 201: If there's no reply, might be blocked inside recv_header(),  so clean_timeout_tasks() might not have chance to trigger. Use a custom Future or select! to achieve this goal.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,452 @@
...
+            received_keepalive_seq: AtomicU64::new(0),
+            packets_keeper: UnsafeCell::new(PacketsKeeper::new(timeout_options.clone().idle_timeout.as_secs())),
+            resp_buf: UnsafeCell::new(BytesMut::with_capacity(8 * 1024 * 1024)),
+        }
+    }
+
+    /// Recv request header from the stream
+    pub async fn recv_header(&self) -> Result<RespHeader, RpcError<String>> {
+        // Try to read to buffer
+        match self.recv_len(REQ_HEADER_SIZE).await {
```
frostyplanet, line 60: recv header at client-side should not use read_exact_timeout, because it depends on whether there's packet sent to server.  (meanswhile  recv body should use read timeout)

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Receive loop for the client.
+    pub async fn recv_loop(&self) {
+        // Check and clean keeper timeout unused task
+        // The timeout data will be cleaned by the get function
+        // We don't need to clean the timeout data radically
+        let mut tickers = tokio::time::interval(self.timeout_options.idle_timeout * 100);
+        loop {
+            // Clean timeout tasks, will block here
+            tokio::select! {
+                _ = tickers.tick() => {
```
frostyplanet, line 193: tokio::select! with only one branch, not sure the ture behavior, unlikely work as exepected.

frostyplanet (changes requested):

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Get the packet status
+    fn status(&self) -> PacketStatus;
+
+    /// Set the packet status
+    fn set_status(&mut self, status: PacketStatus);
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum PacketStatus {
```
frostyplanet, line 143: This can be replace with Option<Result<(), ErrorType>>

src/storage/cache/rpc/client.rs
IRONICBo: Ok, I have removed it.

src/storage/cache/rpc/client.rs
IRONICBo: Thank you for your suggestion, not receiving the result of a recv for a long time can lead to an exception determination timeout.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, fixed.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, I have add a `Debug` trait and add `self` debug.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, fixed.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,283 @@
...
+        if let Some(packet) = self.packets.remove(&seq) {
+            self.timestamp.remove(&seq);
+            return Some(packet);
+        }
+
+        None
+    }
+
+    /// Get a task from the packets
+    pub async fn take_task_mut(&mut self, seq: u64) -> Option<&mut P> {
```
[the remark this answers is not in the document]
IRONICBo: Ok, renamed to `get`.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,283 @@
...
+            packets: HashMap::new(),
+            timestamp: HashMap::new(),
+            timeout,
+        };
+
+        keeper
+    }
+
+    /// Add a task to the packets
+    pub async fn add_task(&mut self, packet: P) {
```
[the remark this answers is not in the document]
IRONICBo: Ok, removed.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,283 @@
...
+        let seq = packet.seq();
+        self.packets.insert(seq, packet);
+        // Get current timestamp
+        // TODO: use a global atomic ticker(updated by check_loop) or read current time?
+        let timestamp = tokio::time::Instant::now().elapsed().as_secs();
+        self.timestamp.insert(seq, timestamp);
+    }
+
+    /// Clean pending tasks as timeout
+    pub async fn clean_timeout_tasks(&mut self) {
```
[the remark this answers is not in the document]
IRONICBo: Ok, removed.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, It's really a problem.
But `recv_header` is wrapped by `read_exact_timeout`, so it's a sync function here.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, I have fixed it.

src/storage/cache/rpc/client.rs
IRONICBo: So we need to block the read here?

frostyplanet: Yes, block until Client is dropped or stream error. write a future to check the closed channel

IRONICBo: If we keep blocking here, it may affect the scheduling of tokio on the client side and cause other tasks not to be called normally.

Recv is similar to try, and will be completed by loop {}.

In addition, the timeout for recv is long enough for the current length of the packet to be read.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,663 @@
...
+    /// Options for the timeout of the connection
+    timeout_options: ClientTimeoutOptions,
+    /// Stream auto increment sequence number, used to mark the request and response
+    seq: AtomicU64,
+    /// Received keep alive seq number, will mark the valid keep alive response.
+    /// If the keep alive response is not received in right mestamp, the connection will be closed,
+    /// If we receive other response, we will update the received_keepalive_timestamp too, just treat it as a keep alive response.
+    received_keepalive_timestamp: AtomicU64,
+    /// Send packet task
+    packets_keeper: UnsafeCell<PacketsKeeper<P>>,
```
frostyplanet, line 41: Remember to change packets_keeper to Mutex

IRONICBo: Thanks, in `PacketsKeeper` inner, it has a RwLock to make sure race safety.

IRONICBo:
```bash
running 13 tests
test storage::cache::rpc::packet::tests::test_req_header_encode_and_decode ... ok
test storage::cache::rpc::packet::tests::test_resp_header_encode_and_decode ... ok
test storage::cache::rpc::client::tests::test_new_connection ... ok
test storage::cache::rpc::client::tests::test_send_data ... ok
test storage::cache::rpc::client::tests::test_next_seq ... ok
test storage::cache::rpc::client::tests::test_ping ... ok
test storage::cache::rpc::client::tests::test_recv_header ... ok
test storage::cache::rpc::client::tests::test_recv_len ... ok
test storage::cache::rpc::workerpool::tests::benchmark_worker_pool ... ok
test storage::cache::rpc::workerpool::tests::test_worker_pool ... ok
test storage::cache::rpc::server::tests::test_rpc_server ... ok
test storage::cache::rpc::tests::test_send_and_recv_packet ... ok
test storage::cache::rpc::packet::tests::test_packets_keeper ... ok

test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 168 filtered out; finished in 13.00s
```

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,524 @@
...
+}
+
+/// The `PacketsInner` struct is used to store the current tasks.
+#[derive(Debug)]
+struct PacketsInner<P>
+where
+    P: Packet + Send + Sync,
+{
+    /// current tasks, marked by the seq number
+    packets: HashMap<u64, P>,
```
frostyplanet, line 196: These two hashmap can be merged into one to reduce operation cost

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,524 @@
...
+    /// other is defined by the user
+    pub op: u8,
+    /// The length of the request.
+    pub len: u64,
+}
+
+impl Encode for ReqHeader {
+    fn encode(&self) -> Vec<u8> {
+        let mut buf = BytesMut::with_capacity(u64_to_usize(REQ_HEADER_SIZE));
+        buf.put_u64(self.seq.to_be());
```
IRONICBo, line 49: to_le()

IRONICBo: receive buffer pointer from caller.

src/storage/cache/rpc/workerpool.rs
```diff
@@ -0,0 +1,242 @@
...
+        Self {
+            max_workers,
+            max_waiting_jobs,
+            job_sender: Arc::new(job_sender),
+            worker_queue,
+        }
+    }
+
+    /// Submit a job to the worker pool asynchronously, and return a future that resolves when the job is completed.
+    pub fn submit_job(&self, job: JobImpl) -> Result<(), RpcError<String>> {
```
rogercloud, line 72: Rename it to `try_submit_job`

src/storage/cache/rpc/workerpool.rs
```diff
@@ -0,0 +1,242 @@
...
+            .try_send(job)
+            .map_err(|_foo| RpcError::InternalError("Failed to submit job".to_owned()))?;
+
+        Ok(())
+    }
+
+    /// Submit a job to the worker pool synchronously, and block until the job is completed.
+    /// Other process will be blocked until the job is completed.
+    /// If all job try to submit the job, the process will be blocked.
+    pub fn submit_job_wait(&self, job: JobImpl) -> Result<(), RpcError<String>> {
```
rogercloud, line 89: Rename it to `submit_job`.

src/storage/cache/rpc/workerpool.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub async fn wait_for_completion(&self) {
+        // Check the sender channel is empty and all workers are completed.
+        while !self.job_sender.is_empty() {
+            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+        }
+    }
+
+    /// Delete the worker pool and shutdown all workers.
+    pub fn shutdown(&self) {
+        debug!("Shutting down worker pool...");
```
rogercloud, line 108: Not finished yet?

src/storage/cache/rpc/workerpool.rs
```diff
@@ -0,0 +1,242 @@
...
+            .send(job)
+            .map_err(|_foo| RpcError::InternalError("Failed to submit job".to_owned()))?;
+
+        Ok(())
+    }
+
+    /// Wait for all workers to complete.
+    pub async fn wait_for_completion(&self) {
+        // Check the sender channel is empty and all workers are completed.
+        while !self.job_sender.is_empty() {
```
rogercloud, line 101: I don't think the empty of the channel means that all workers are completed.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok.

src/storage/cache/rpc/utils.rs
```diff
@@ -0,0 +1,81 @@
...
+        .ok_or_else(|| RpcError::InternalError("Invalid range data".to_owned()))
+        .map(|&byte| u8::from_be(byte))
+}
+
+/// Converts [`u64`] to [`usize`], and panic when the conversion fails.
+#[allow(clippy::missing_const_for_fn)] // <- false positive
+#[allow(clippy::expect_used)] // We can ensure that this method won't panic if we followed the hints above
+#[inline]
+#[must_use]
+pub fn u64_to_usize(x: u64) -> usize {
```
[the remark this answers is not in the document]
IRONICBo: converted to const.

src/storage/cache/rpc/workerpool.rs
IRONICBo: Yes, the check here is just to make sure that all the jobs are consumed, we don't care about the success or failure of the task here, the caller can use mechanisms such as channel inside the job to determine if the job is complete or not.

I've modify the comments.

src/storage/cache/rpc/workerpool.rs
IRONICBo: This is the legacy message for debug, all workers and channels are automatically dropped by `Drop`.

src/storage/cache/rpc/workerpool.rs
IRONICBo: OK.

src/storage/cache/rpc/workerpool.rs
IRONICBo: OK.

src/storage/cache/rpc/packet.rs
IRONICBo: I've add a `set_timestamp` for `Packet`, so we don't need another map.

src/storage/cache/rpc/server.rs
```diff
@@ -0,0 +1,478 @@
...
+/// The worker factory for the RPC connection.
+#[derive(Clone, Debug)]
+pub struct RpcConnWorkerFactory<T>
+where
+    T: RpcServerConnectionHandler + Send + Sync + 'static,
+{
+    /// Global worker pool for the RPC connection, shared by all connections.
+    worker_pool: Arc<WorkerPool>,
+    /// The handler for the connection
+    dispatch_handler: T,
```
frostyplanet, line 297: dispatch_handler is not use by the struct, there will be warnning?

IRONICBo: Yes, it is better to be removed.

src/storage/cache/rpc/workerpool.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self {
+            max_workers,
+            max_waiting_jobs,
+            job_sender: Arc::new(job_sender),
+            worker_queue,
+        }
+    }
+
+    /// Submit a job to the worker pool asynchronously, and return a future that resolves when the job is completed.
+    pub fn try_submit_job(&self, job: JobImpl) -> Result<(), RpcError<String>> {
```
frostyplanet, line 72: 1. There should not be is try_submit_job() to worker pool,  it has no read usage. Make sure we always use submit_job().
2. Two problem in this function:
a). if is_full and try_send() is not atomic,  try_send() might still failed because of full, since under laying channel is mpsc. 
b). Err should distinguish full error and channel close error.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,525 @@
...
+        self.packets.remove(&seq)
+    }
+
+    /// Clean pending tasks as timeout
+    /// In first step, we will try to mark the task as timeout if it is pending and timeout
+    /// In 10 * timeout time, we will force to remove the task if the task is not consumed
+    pub fn clean_timeout_tasks(&mut self, current_timestamp: u64, timeout: u64) {
+        let mut last_timeout_packets = Vec::new();
+        for (seq, packet) in &mut self.packets {
+            let current_status = packet.status();
```
frostyplanet, line 230: This function has logical problem, it should be one step. ignore the PacketStatus

src/storage/cache/rpc/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+            // check current buffer is req
+            let req = FileBlockRequest::decode(&self.buffer)?;
+
+            // Return a 4MB vec
+            Ok(req.encode())
+        }
+
+        fn set_resp_data(&mut self, data: &[u8]) -> Result<(), RpcError<String>> {
+            // Try to set the response data
+            self.buffer.clear();
```
frostyplanet, line 280: Why not decode directly and store the structured data, instead of [u8] ?

frostyplanet (changes requested):

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,524 @@
...
+            let seq = packet.seq();
+            packets_inner.add_task(seq, timestamp, packet);
+        }
+
+        let current_timestamp = self.current_time.elapsed().as_secs();
+        packets_inner.clean_timeout_tasks(current_timestamp, self.timeout);
+    }
+
+    /// Get a task from the packets, and update data in the task
+    pub fn update_task_mut(
```
frostyplanet, line 337: It'd better rename update_task_mut(&mut self)->  take_task(&self)

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,643 @@
...
+                req_packet.seq()
+            );
+            Err(RpcError::InternalError(
+                "Failed to serialize request".to_owned(),
+            ))
+        }
+    }
+
+    /// Receive packet by the client
+    pub fn recv_packet(&self, resp_packet: &P) -> Option<P> {
```
frostyplanet, line 223: Unused function, delete it

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,524 @@
...
+                RpcError::InternalError(format!("Failed to send packet to buffer: {e:?}"))
+            })?;
+
+        Ok(())
+    }
+
+    /// Consume a task from inner
+    /// If return task is none, means that current task is in buffer,
+    /// and haven't triggered timeout or get response in time
+    pub fn consume_task(&mut self, seq: u64) -> Option<P> {
```
frostyplanet, line 318: Unused function, delete it,  (duplicated of update_task_mut)

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,652 @@
...
+        packets_keeper.consume_task(resp_packet.seq())
+    }
+
+    /// Receive loop for the client.
+    pub async fn recv_loop(&self) {
+        // Check and clean keeper timeout unused task
+        // The timeout data will be cleaned by the get function
+        // We don't need to clean the timeout data radically
+        let mut tickers = Box::pin(tokio::time::interval(
+            self.timeout_options.idle_timeout * 100,
```
frostyplanet, line 235: idle_timeout should be rename to task_timeout.

ticker should be task_timeout / 2 ,   if ticker > task_timeout,   the actual timeout will be inacurate

IRONICBo: idle_timeout to task_timeout

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,652 @@
...
+
+        fn get_resp_data(&self) -> Result<Vec<u8>, RpcError<String>> {
+            Ok(Vec::new())
+        }
+
+        fn status(&self) -> PacketStatus {
+            self.status
+        }
+
+        fn set_status(&mut self, status: PacketStatus) {
```
frostyplanet, line 522: change to set_result(self,  Result<resp_data, Err()>)

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,652 @@
...
+
+        fn get_req_data(&self) -> Result<Vec<u8>, RpcError<String>> {
+            Ok(Vec::new())
+        }
+
+        fn set_resp_data(&mut self, _data: &[u8]) -> Result<(), RpcError<String>> {
+            Ok(())
+        }
+
+        fn get_resp_data(&self) -> Result<Vec<u8>, RpcError<String>> {
```
frostyplanet, line 514: get_resp_data(&self) -> Result<(), RpcError>

src/storage/cache/rpc/error.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Error types for the RPC server and client
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum RpcError<T> {
+    /// The request is invalid.
+    InvalidRequest(T),
+    /// The response is invalid.
+    InvalidResponse(T),
+    /// The server/client meet an internal error.
+    InternalError(T),
+}
```
frostyplanet, line 12: It would be better remote T from generic,  since all the places use String.
add Timeout for convienent

src/storage/cache/rpc/error.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl<T> RpcError<T> {
+    /// Convert the error to a string.
+    pub fn to_string(self) -> T {
```
frostyplanet, line 16: replace usage with Display for Debug

src/storage/cache/rpc/workerpool.rs
IRONICBo: Ok, I have removed this function.

src/storage/cache/rpc/packet.rs
IRONICBo: This was originally designed to be a delayed deletion, now the packet notifies the caller immediately that the behaviour has been deleted.

src/storage/cache/rpc/mod.rs
IRONICBo: We can do decode inside set_resp_data or set_result, passing in buffer here just avoids putting the decode logic into the rpc (decode requires type).

src/storage/cache/rpc/packet.rs
IRONICBo: OK.

src/storage/cache/rpc/client.rs
IRONICBo: OK, update to use channel to sync packet.

src/storage/cache/rpc/packet.rs
IRONICBo: OK.

src/storage/cache/rpc/client.rs
IRONICBo: Changed to `self.timeout_options.task_timeout.div_f32(2.0)`.

src/storage/cache/rpc/client.rs
IRONICBo: resp_data is in `set_resp_data`, changed to `        fn set_result(&self, _status: Result<(), RpcError>) {}`.

src/storage/cache/rpc/client.rs
IRONICBo: I've deleted this function.

src/storage/cache/rpc/error.rs
IRONICBo: OK.

src/storage/cache/rpc/error.rs
IRONICBo: OK, I have removed it.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+
+    #[derive(Debug, Clone)]
+    pub struct TestPacket {
+        pub seq: u64,
+        pub op: u8,
+        pub timestamp: u64,
+    }
+
+    impl Encode for TestPacket {
+        fn encode(&self, _buf: &mut BytesMut) {}
```
frostyplanet, line 458: Should implement this, aotherwise the behavior is not tested

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+        }
+
+        fn set_resp_data(&mut self, _data: &[u8]) -> Result<(), RpcError> {
+            Ok(())
+        }
+
+        fn set_result(&self, _status: Result<(), RpcError>) {}
+
+        fn get_req_len(&self) -> u64 {
+            0
```
frostyplanet, line 493: Should implement this

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,502 @@
...
+    fn set_op(&mut self, op: u8);
+
+    /// Set timestamp
+    fn set_timestamp(&mut self, timestamp: u64);
+    /// Get timestamp
+    fn get_timestamp(&self) -> u64;
+
+    /// Get request buf length
+    /// This function is used to get the length of the request size for join header
+    fn get_req_len(&self) -> u64;
```
frostyplanet, line 137: Should return only the len of message, not the header,  because this len filled into Header, receiver received the fixed length header to discover how many body message pending to receive

frostyplanet (changes requested):

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+        tokio::spawn(async move {
+            inner_connection_clone.recv_loop().await;
+        });
+    }
+
+    /// Send a request to the server.
+    /// Try to send data to channel, if the channel is full, return an error.
+    /// Contains the rezquest header and body.
+    /// WARN: this function does not support concurrent call
+    pub async fn send_request(&self, req: &mut P) -> Result<(), RpcError> {
```
frostyplanet, line 363: `req: &mut P` should change to `req: mut P`

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+        } else {
+            debug!("{:?} Failed to send keep alive message", self);
+            Err(RpcError::InternalError(
+                "Failed to send keep alive message".to_owned(),
+            ))
+        }
+    }
+
+    /// Send packet by the client.
+    pub async fn send_packet(&self, req_packet: &mut P) -> Result<(), RpcError> {
```
frostyplanet, line 197: `req_packet: &mut P `  change to `req_packet: mut P`

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+            len: req_len,
+        };
+
+        // concate req_header and req_buffer
+        if let Ok(()) = self.send_data(&req_header, Some(req_packet)).await {
+            debug!("{:?} Sent request success: {:?}", self, req_packet.seq());
+            // We have set a copy to keeper and manage the status for the packets keeper
+            // Set to packet task with clone
+            self.get_packets_keeper_mut()
+                .add_task(&mut req_packet.clone())?;
```
frostyplanet, line 216: Do not clone, move the packet inside

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+            inner: packets_inner,
+            buffer_packets_sender,
+            buffer_packets_receiver,
+            timeout,
+            current_time,
+        }
+    }
+
+    /// Add a task to the packets
+    pub fn add_task(&mut self, packet: &mut P) -> Result<(), RpcError> {
```
frostyplanet, line 253: Should be `packet: mut P`

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+        // TODO: use a global atomic ticker(updated by check_loop) or read current time?
+        let timestamp = self.current_time.elapsed().as_secs();
+        packet.set_timestamp(timestamp);
+        self.buffer_packets_sender
+            .send(packet.clone())
```
frostyplanet, line 258: Remember when dealing with channel message, do not clone.
Take the packet return in SendError, set proper error to it,  otherwise might lead to hang if channel can be closed

frostyplanet (changes requested):

src/storage/cache/rpc/packet.rs
frostyplanet: Remove mut from self, no longer need it.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+            .map_err(|e| {
+                RpcError::InternalError(format!("Failed to send packet to buffer: {e:?}"))
+            })?;
+
+        Ok(())
+    }
+
+    /// Clean pending tasks as timeout
+    pub fn clean_timeout_tasks(&mut self) {
+        let mut packets_inner = self.inner.write();
```
frostyplanet, line 268: Remove mut from self, no longer need it.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+            let seq = packet.seq();
+            packets_inner.add_task(seq, packet);
+        }
+
+        let current_timestamp = self.current_time.elapsed().as_secs();
+        packets_inner.clean_timeout_tasks(current_timestamp, self.timeout);
+    }
+
+    /// Purge the outdated tasks when connection is timeout
+    pub fn purge_outdated_tasks(&mut self) {
```
frostyplanet, line 279: Remove mut from self, no longer need it.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,640 @@
...
+    /// Stream auto increment sequence number, used to mark the request and response
+    seq: AtomicU64,
+    /// The instant of the connection, used to calculate the keep alive timeout
+    clock_instant: Instant,
+    /// Received keep alive timestamp, will mark the valid keep alive response.
+    /// If the keep alive response is not received in right mestamp, the connection will be closed,
+    /// If we receive other response, we will update the received_keepalive_timestamp too, just treat it as a keep alive response.
+    received_keepalive_timestamp: AtomicU64,
+    /// Send packet task
+    packets_keeper: UnsafeCell<PacketsKeeper<P>>,
```
frostyplanet, line 46: packets_keeper no longer need UnsafeCell

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+    struct TestRequest {
+        pub mock: u64,
+    }
+
+    impl Encode for TestRequest {
+        fn encode(&self, _buf: &mut BytesMut) {}
+    }
+
+    #[derive(Debug, Clone)]
+    struct TestPacket {
```
frostyplanet, line 347: Instead of Define TestPacket, you can directly use the Packet from message.rs,  to save some coding

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl Decode for FileBlockRequest {
+    /// Decode the byte buffer into a file block request.
+    fn decode(buf: &[u8]) -> Result<Self, RpcError> {
+        decode_file_block_request(buf)
+    }
+}
+
+/// Decode the file block request from the buffer.
+pub fn decode_file_block_request(buf: &[u8]) -> Result<FileBlockRequest, RpcError> {
```
frostyplanet, line 99: The decode & encode better be member function.
Should add some test for encode and then decode to valid the behavior

frostyplanet (changes requested):

src/storage/cache/rpc/client.rs
IRONICBo: OK, The encode logic here is mostly placed in the overall test, so it's simplified here, but I'll still add this.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, I have add the real length of the request.

src/storage/cache/rpc/packet.rs
IRONICBo: Yes, actually this function only return the length of the request.
This length will be filled in header's struct(length):
```rs
let req_len = req_packet.get_req_len(); // Try to get request buffer
let req_header = ReqHeader {
    seq: req_packet.seq(),
    op: req_packet.op(),
    len: req_len,
};
```

src/storage/cache/rpc/client.rs
IRONICBo: Ok

src/storage/cache/rpc/client.rs
IRONICBo: Ok.

IRONICBo: It seems that function does not support mut P.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, I have removed the clone function.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, I have changed it.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, I have changed it to `to_owned`.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, removed.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok.

src/storage/cache/rpc/client.rs
IRONICBo: Maybe it is needed to be modified in `ReceiveHeaderFuture`.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok.

src/storage/cache/rpc/message.rs
IRONICBo: Ok, I have add the unit tests.

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,322 @@
...
+                        match self.sender.send(Ok(file_block_resp)) {
+                            Ok(()) => {}
+                            Err(err) => {
+                                error!("Failed to set result: {:?}", err);
+                            }
+                        }
+                    }
+                    None => {
+                        error!("Failed to set result: response is None");
+                        match self.sender.send(Err(self.request.clone())) {
```
frostyplanet, line 302: Note that here use send() which might block,  you should make sure sender is from Unbounded channel ,  but flume does not provide trait to ensure that.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,500 @@
...
+    timeout: u64,
+    /// current timestamp
+    current_time: Instant,
+}
+
+impl<P: Packet + Send + Sync> PacketsKeeper<P> {
+    /// Create a new `PacketsKeeper`
+    #[must_use]
+    pub fn new(timeout: u64) -> Self {
+        let (buffer_packets_sender, buffer_packets_receiver) = flume::bounded::<P>(1000);
```
frostyplanet, line 239: Should either change it to unbound or use send_async/recv_async

src/storage/cache/rpc/message.rs
IRONICBo: Yes, the use of blocking and non-blocking code at runtime in tokio test may affect the logic of the implementation, have switched the relevant to async implementation.

src/storage/cache/rpc/packet.rs
IRONICBo: Ok, fixed.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,660 @@
...
+        }
+
+        let buffer: &mut BytesMut = unsafe { &mut *self.resp_buf.get() };
+        let req_header = RespHeader::decode(buffer)?;
+        debug!("Received response header: {:?}", req_header);
+
+        Ok(req_header)
+    }
+
+    /// Receive request body from the server.
```
rogercloud, line 107: Is the comment correct?

src/storage/cache/rpc/common.rs
```diff
@@ -0,0 +1,46 @@
...
+
+/// Options for the timeout of the connection
+#[derive(Debug, Clone)]
+pub struct ClientTimeoutOptions {
+    /// The timeout for reading data from the connection
+    pub read_timeout: Duration,
+    /// The timeout for writing data to the connection
+    pub write_timeout: Duration,
+    /// The timeout for the idle connection
+    pub task_timeout: Duration,
```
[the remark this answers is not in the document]
IRONICBo: This is a suggestion provided by @frostyplanet , as the timeout here is mainly for single packet (task) timeout configurations, and the note here may not have been updated.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,660 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+impl<P> RpcClientConnectionInner<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    /// Create a new connection.
+    pub fn new(stream: TcpStream, timeout_options: &ClientTimeoutOptions, client_id: u64) -> Self {
```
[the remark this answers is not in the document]
IRONICBo: Okay, we can internally construct a timeout struct.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,660 @@
...
+        Self {
+            stream: UnsafeCell::new(stream),
+            timeout_options: timeout_options.clone(),
+            seq: AtomicU64::new(0),
+            clock_instant: Instant::now(),
+            received_keepalive_timestamp: AtomicU64::new(0),
+            packets_keeper: PacketsKeeper::new(timeout_options.clone().task_timeout.as_secs()),
+            resp_buf: UnsafeCell::new(BytesMut::with_capacity(8 * 1024 * 1024)),
+            req_buf: UnsafeCell::new(BytesMut::with_capacity(8 * 1024 * 1024)),
```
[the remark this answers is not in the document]
IRONICBo: Ok.

src/storage/cache/rpc/client.rs
IRONICBo: It should be response body, I will modify it.

src/storage/cache/rpc/common.rs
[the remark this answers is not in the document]
frostyplanet: @IRONICBo  on server side there should be idle_timeout to close idle (maybe dead) connection (task timeout is not relevant to server)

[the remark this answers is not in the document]
IRONICBo: OK, I will update it.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+                Err(RpcError::InternalError(err.to_string()))
+            }
+        }
+    }
+
+    /// Send a request to the server.
+    /// We need to make sure
+    /// Current send packet need to be in one task, so if we need to send ping and packet
+    /// we need to make sure only one operation is running, like select.
+    pub async fn send_data(
```
frostyplanet, line 134: Please remove all the `pub` in methods of `RpcClientConnectionInner`

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+            op: req_packet.op(),
+            len: req_len,
+        };
+
+        // concate req_header and req_buffer
+        if let Ok(()) = self.send_data(&req_header, Some(req_packet)).await {
+            debug!("{:?} Sent request success: {:?}", self, req_packet.seq());
+            // We have set a copy to keeper and manage the status for the packets keeper
+            // Set to packet task with clone
+            self.packets_keeper.add_task(req_packet)?;
```
frostyplanet, line 217: Move add_task() inside send_data

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+    timeout_options: ClientTimeoutOptions,
+    /// Inner connection for the client.
+    inner_connection: Arc<RpcClientConnectionInner<P>>,
+    /// Client ID for the connection
+    /// TODO:
+    #[allow(dead_code)]
+    client_id: AtomicU64,
+}
+
+impl<P> RpcClient<P>
```
frostyplanet, line 317: You should impl Drop for `RpcClient`, gracefully shutdown sender part, and notify recv_loop() to exit

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+                        debug!("{:?} Invalid response type: {:?}", self, header.op);
+                        break;
+                    }
+                }
+                Err(err) => {
+                    debug!("{:?} Failed to receive request header: {:?}", self, err);
+                    break;
+                }
+            }
+        }
```
frostyplanet, line 284: Remember to purge_outdated_tasks() before recv_loop() exit

src/storage/cache/rpc/server.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let buffer: &mut BytesMut = unsafe { &mut *self.req_buf.get() };
+        let req_header = ReqHeader::decode(buffer)?;
+        debug!("Received request header: {:?}", req_header);
+
+        Ok(req_header)
+    }
+
+    /// Recv request body from the stream
+    pub async fn recv_len(&self, len: u64) -> Result<(), RpcError> {
```
frostyplanet, line 127: Should add an timeout arg to recv_len().
use idle_timeout for header ,  and use read_timeout for body

src/storage/cache/rpc/server.rs
```diff
@@ -0,0 +1,486 @@
...
+        }
+    }
+
+    /// Keep the connection and get the handler for the connection.
+    pub async fn run(&self) {
+        // Dispatch the handler for the connection
+        debug!("RpcServerConnection::Received a new TcpStream.");
+
+        // TODO: copy done_tx to the worker pool
+        let (done_tx, mut done_rx) = mpsc::channel::<Vec<u8>>(10000);
```
frostyplanet, line 245: Should use unbound channel to avoid blocking set_result()

frostyplanet (changes requested):

src/storage/cache/rpc/client.rs
IRONICBo: OK, I added a StreamReaderFuture to check client and stream is closed.

src/storage/cache/rpc/client.rs
IRONICBo: OK.

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,440 @@
...
+/// The request type of the request.
+#[derive(Debug)]
+pub enum ReqType {
+    /// The keep alive request.
+    KeepAliveRequest,
+    /// The file block request.
+    FileBlockRequest,
+}
+
+impl ReqType {
```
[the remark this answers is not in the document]
IRONICBo: OK, so I will update this mapping with `https://github.com/illicitonion/num_enum`.

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,440 @@
...
+/// The operation type of the response.
+#[derive(Debug)]
+pub enum RespType {
+    /// The keep alive response.
+    KeepAliveResponse,
+    /// The file block response.
+    FileBlockResponse,
+}
+
+impl RespType {
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,440 @@
...
+    fn decode(buf: &[u8]) -> Result<Self, RpcError> {
+        if buf.len() < 41 {
+            return Err(RpcError::InternalError("Insufficient bytes".to_owned()));
+        }
+        let file_id = get_u64_from_buf(buf, 0)?;
+        let block_id = get_u64_from_buf(buf, 8)?;
+        let block_size = get_u64_from_buf(buf, 16)?;
+        let block_version = get_u64_from_buf(buf, 24)?;
+        let hash_ring_version = get_u64_from_buf(buf, 32)?;
+        let status = match buf.get(40) {
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/storage/cache/rpc/message.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub response: Option<FileBlockResponse>,
+    /// The sender to send response back to caller.
+    pub done_tx: flume::Sender<Result<FileBlockResponse, FileBlockRequest>>,
+}
+
+impl FileBlockPacket {
+    /// Create a new file block packet.
+    #[must_use]
+    pub fn new(
+        block_request: &FileBlockRequest,
```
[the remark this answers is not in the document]
IRONICBo: OK, remove `&` reference mark.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,514 @@
...
+        // Mark timeout
+        let packet_2 = TestPacket {
+            seq: 2,
+            op: 2,
+            timestamp: 0,
+            buf: BytesMut::new(),
+            sender: tx.clone(),
+            request: TestRequest { mock: 0 },
+        };
+        packets_keeper.add_task(&mut packet_2.clone()).unwrap();
```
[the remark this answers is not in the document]
IRONICBo: OK, I will update this test usage.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,514 @@
...
+        // Purge the timeout task
+        let packet_3 = TestPacket {
+            seq: 3,
+            op: 2,
+            timestamp: 0,
+            buf: BytesMut::new(),
+            sender: tx.clone(),
+            request: TestRequest { mock: 0 },
+        };
+        packets_keeper.add_task(&mut packet_3.clone()).unwrap();
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/storage/cache/rpc/packet.rs
```diff
@@ -0,0 +1,514 @@
...
+        // Success
+        let packet = TestPacket {
+            seq: 1,
+            op: 2,
+            timestamp: 0,
+            buf: BytesMut::new(),
+            sender: tx.clone(),
+            request: TestRequest { mock: 0 },
+        };
+        packets_keeper.add_task(&mut packet.clone()).unwrap();
```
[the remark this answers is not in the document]
IRONICBo: OK.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+
+        let buffer: &mut BytesMut = unsafe { &mut *self.resp_buf.get() };
+        let req_header = RespHeader::decode(buffer)?;
+        debug!("Received response header: {:?}", req_header);
+
+        Ok(req_header)
+    }
+
+    /// Receive response body from the server.
+    pub async fn recv_len(&self, len: u64) -> Result<(), RpcError> {
```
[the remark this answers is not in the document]
IRONICBo: This comment means that we shoud not use timeout here? 
@frostyplanet  https://github.com/datenlord/datenlord/pull/523#discussion_r1740205978

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,665 @@
...
+        let self_mut: &mut ReceiveHeaderFuture<'a, P, F> = self.get_mut();
+        let client_inner = self_mut.client_inner;
+
+        // clean timeout tasks
+        let mut tick_ready = false;
+        while self_mut.interval.as_mut().poll_tick(cx).is_ready() {
+            // consume all ticks in once call
+            tick_ready = true;
+        }
+        while tick_ready {
```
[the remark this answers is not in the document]
IRONICBo: OK, so should I split this loop to small batch?

src/storage/cache/rpc/client.rs
IRONICBo: OK.

src/storage/cache/rpc/client.rs
IRONICBo: OK, add a `add_task()` function for this client.

src/storage/cache/rpc/client.rs
IRONICBo: OK.

src/storage/cache/rpc/client.rs
IRONICBo: OK, same as `gracefully shutdown` here.

src/storage/cache/rpc/server.rs
IRONICBo: OK, that make sense for me.

src/storage/cache/rpc/server.rs
IRONICBo: OK, remove `1000` capacity here.

frostyplanet: https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.unbounded_channel.html

src/storage/cache/rpc/client.rs
IRONICBo: Modified to `mut req: P`.

src/storage/cache/rpc/client.rs
IRONICBo: Changed to `pub async fn send_packet(&self, mut req_packet: P)`.

src/storage/cache/rpc/client.rs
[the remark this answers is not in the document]
IRONICBo: @frostyplanet

src/storage/cache/rpc/client.rs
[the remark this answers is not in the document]
IRONICBo: clean_timeout_tasks  need to call `set_result` and use channel's `send_async` function, so we may need to keep the `async` ?

src/storage/cache/rpc/client.rs
IRONICBo: Maybe we don't want to keep ping message to packets_keeper?

IRONICBo: In `send_data`, it only have `Encode` trait not the packet trait.

Encode trait is for `header` and `body`, but packets_keeper only keep the body.

src/storage/cache/rpc/client.rs
IRONICBo: `RecvLenFuture`

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,760 @@
...
+
+/// Drop the client connection, send the close signal to the client
+/// Other side will receive the close signal and close the connection
+impl<P> Drop for RpcClientConnectionInner<P>
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    fn drop(&mut self) {
+        // Send the close signal to the client
+        match self.is_closed_sender.send(()) {
```
frostyplanet, line 80: impl drop for RpcClientConnectionInner  is not effected,  since RpcClientConnectionInner is hold by recv_loop. 
Delete the drop code,  move is_closed_sender to RpcClient,  receiver get notification when is_closed_sender is dropped.

src/storage/cache/rpc/client.rs
```diff
@@ -0,0 +1,760 @@
...
+where
+    P: Packet + Clone + Send + Sync + 'static,
+{
+    type Output = Result<(), RpcError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+
+        // Recv closed data from the client channel
+        if this.is_closed.try_recv().is_ok() {
```
frostyplanet, line 440: You should poll the is_closed channel,  using try_recv() will not be notified to wake up

frostyplanet (changes requested):

src/storage/cache/rpc/client.rs
IRONICBo: Ok, move sender to `RpcClient` and init inner with receiver.

src/storage/cache/rpc/client.rs
IRONICBo: Ok, this struct will use `recv_async` future and use `poll` to reach channel status.

src/storage/cache/rpc/client.rs
```diff
@@ -297,8 +297,32 @@ where
...
+    /// Get resp buffer with mutable reference
+    #[allow(clippy::mut_from_ref)]
+    fn get_resp_buf_mut(&self) -> &mut BytesMut {
+        unsafe { std::mem::transmute::<*mut BytesMut, &mut BytesMut>(self.resp_buf.get()) }
+    }
+
+    /// Get req buffer with mutable reference
+    #[allow(clippy::mut_from_ref)]
+    fn get_req_buf_mut(&self) -> &mut BytesMut {
+        unsafe { std::mem::transmute::<*mut BytesMut, &mut BytesMut>(self.req_buf.get()) }
```
frostyplanet, line 312: just   `  unsafe { std::mem::transmute(self.req_buf.get()) }`  will do

IRONICBo: OK, remove the type annotation.

frostyplanet (approved):
LGTM

PR #524, opened 2024-05-31
feat(cache node): support cluster config manager.

IRONICBo:
# Background
At present, the core functionality and core logic of each component of distributed cache has been completed design, and the core code has also been submitted to the cache branch, so now we need to do the upper layer manager part of the package for the original requirements of the cache node, in order to provide distributed caching capabilities in the intranet environment.
We need to design the upper layer manager for intranet environments where cache node changes are not very frequent.

Our design mainly focuses on the block management/block distribution of cache nodes, i.e., to provide block caching capability on the server side and cluster management of server nodes; in addition, to provide a module or a lib on the client side to be integrated into the existing storage for accessing nodes via hashring, as well as to provide the client with the ability to access nodes via In addition, on the client side, we provide a module or lib to be integrated into the existing storage to access the node via hashring, and to set up an rpc client to fetch the blocks, as well as the necessary error handling.

# Design
1. Cluster
This part of the design in the cluster already has a more detailed process, in the manager side just need to run up the FSM can be
2. Cache
 Cache design provides a load/store interface to mem design.
3. RPC
In the server side need to implement the trait provided by rpc, need to implement the logic of local cache and back to origin
4. Config Private configuration for a single server
The config part is aligned with the datenlord, so that when reading and writing blocks, it can be aligned with the datenlord.
the config is private to each cluster, not shared by the cluster.
5. ClusterConfig public configuration in cluster
Cluster server subscribes to the public configuration of the cluster, including
block_size changes, i.e., the cluster will respond to changes in block size, which is currently a direct drop of the cache. 2.
Refresh, i.e., manually drop the configuration in the cache, which is used by operation and maintenance.

Ref: https://datenlord.feishu.cn/docx/HufOdyeffocslXxP6SpcrHY0nhc


frostyplanet:
The last one to review

PR #521, opened 2024-05-08
fix(fuse): unexcepted meta data delete.

IRONICBo:
On a standalone node we use forget in order to receive messages sent by the fuse kernel to support open-to-close, but in distributed scenarios there may be a risk of accidentally deleting meta data, and there is a need to limit the functionality of forget as well as remove the forget capability.

PR #525, opened 2024-06-04
fix: using environment variables instead of moving files

IRONICBo:
Current `cron-test` need to install kind with root, we want to install kind executable file and export to environment variables.


IRONICBo:
Please do not merge it after https://github.com/datenlord/datenlord/pull/521 is fixed.

scripts/setup/install-kind.sh
```diff
@@ -23,4 +25,5 @@ fi
 echo "Installing kind"
 curl -Lo ./kind https://kind.sigs.k8s.io/dl/v$KIND_VERSION/kind-linux-amd64
 chmod +x ./kind
-sudo mv ./kind /usr/local/bin/kind
\ No newline at end of file
+echo "kind is in $PATH"
```
IRONICBo, line 28: Move to `$HOME/bin`.

frostyplanet (approved):

PR #527, opened 2024-06-26
feat(cache node): support cluster manager.

IRONICBo:
# Design 

In this PR, we have implemented the cluster manager for distribute cache nodes, typically for these cases:
- Election Testing: Ensure that exactly one node is elected as the control node.
- Node Addition Testing: The controller should detect new nodes and modify the hash ring accordingly.
- Offline Logic Testing: Implement testing for node disconnections, e.g., by aborting coroutines, allowing etcd to detect these changes.
- Detection and Synchronization: Other nodes should detect changes in the etcd's hash ring and synchronize their in-memory state accordingly.

The core struct is `ClusterManager` & `ClusterManagerInner`.

### ClusterManager

```rs
pub struct ClusterManager {
    /// inner struct
    inner: Arc<ClusterManagerInner>,
}

impl ClusterManager {
    /// Create new cluster manager
    pub fn new(kv_engine: Arc<KVEngineType>, node: Node) -> Self {}
    
    /// Get current node
    #[must_use]
    pub fn get_node(&self) -> Node {}

    /// Get current hashring
    pub async fn get_ring(&self) -> DatenLordResult<Ring<Node>> {}
    
    /// Get all nodes
    pub async fn get_nodes(&self) -> DatenLordResult<Vec<Node>> {}
    
    ///  Stop the cluster manager
    pub fn stop(&self) {}
    
    /// Run the cluster manager as state machine
    ///
    /// We need to perpare current node info, current node list and generate hash ring info
    pub async fn run(&self) -> DatenLordResult<()> {}
    
    /// Do master tasks
    ///
    /// 1. Master node will watch the node list update, and update the ring
    /// 2. Master will check self
    async fn do_master_tasks(&self) -> DatenLordResult<()> {}
    
    /// Do slave tasks
    ///
    /// Slave node will watch the ring update and campaign master
    #[allow(clippy::arithmetic_side_effects, clippy::pattern_type_mismatch)] // The `select!` macro will generate code that goes against these rules.
    async fn do_slave_tasks(&self) -> DatenLordResult<()> {}
}
```

- run() core cluster manager method, i.e. synchronise etcd information/cluster configuration management, corresponds to ClusterManager in module relationship.
- get_xxx() exposes to other modules that hold ClusterManager, which can get information such as hashring maintained by the current node, and is internally managed by ArcSwap, which will return a copy of the current version (there may be some copying, but for the case where the number of nodes is not so large and updates are not frequent, it is a way to simplify the implementation). implementation).
- do_xxx_tasks() was originally placed inside inner, but ownership issues can be difficult to deal with when opening multiple tasks, so it puts the task here in ClusterManager and splits the read/write task inside the task.

### ClusterManagerInner

```rs
pub struct ClusterManagerInner {
    /// Etcd client
    kv_engine: Arc<KVEngineType>,
    /// Node session, try to keep the session alive, we will set the session in run method
    node_session: ArcSwapOption<Session>,
    /// current node info
    node: ArcSwap<Node>,
    /// current node list, we will set the session in run method
    node_list: ArcSwap<Vec<Node>>,
    /// current hash ring, we will set the session in run method
    hash_ring: ArcSwap<Ring<Node>>,
    // TODO: Support distribute cluster config here
}

impl ClusterManagerInner {
    /// Create a new cluster manager
    pub fn new(kv_engine: Arc<KVEngineType>, node: Node) -> Self {}
    
    /// Prepare for the session and other data
    ///
    /// We need to perpare current node info, current node list and generate hash ring info
    pub async fn prepare(&self) -> DatenLordResult<()> {}
    
    /// Update node info
    pub async fn update_node_in_cluster(&self) -> DatenLordResult<()> {}
    
    /// Register current node to etcd and keep alive
    pub async fn register(&self) -> DatenLordResult<()> {}
    
    /// Do campaign
    ///
    /// Try to campaign master, will return status and master value
    pub async fn do_campaign(&self) -> DatenLordResult<(bool, String)> {}
    
    /// Write current cluster info to etcd
    #[allow(clippy::arithmetic_side_effects, clippy::pattern_type_mismatch)] // The `select!` macro will generate code that goes against these rules.
    async fn write_task(&self, rx: flume::Receiver<()>, token: CancellationToken) {}
    
    /// Clean up the tasks
    pub fn clean_sessions(&self) {}
    
    /// Clean up the tasks
    pub fn clean_sessions(&self) {}
    
    /// Master node will watch the node list update, and update the ring
    pub async fn watch_nodes(&self, tx: flume::Sender<()>) -> DatenLordResult<()> {}
    
    /// Update cluster topo
    ///
    /// Try to update cluster by current master node
    async fn update_cluster_topo(&self) -> DatenLordResult<()> {}

    /// Try to watch the master node
    /// If the master node is down, the slave node will try to get the master lock
    /// Then current node will become the master node if operation is successful
    pub async fn watch_master(&self) -> DatenLordResult<()> {}
    
    /// Save ring to etcd
    async fn save_ring(&self) -> DatenLordResult<()> {}
    
    /// Load ring from etcd
    async fn load_ring(&self) -> DatenLordResult<Ring<Node>> {}
    
    /// Get node list from etcd
    async fn load_nodes(&self) -> DatenLordResult<Vec<Node>> {}
    
    /// Get all nodes
    ///
    /// This function is used to get all nodes from etcd
    /// If we set `must_fetch`, we will also update current node list
    pub async fn get_nodes(&self, must_fetch: bool) -> DatenLordResult<Vec<Node>> {}
    
    /// Get current node
    pub fn get_node(&self) -> Node {}
    
    /// Update current node
    pub fn update_node(&self, node: Node) {}
    
    /// Get hashring
    ///
    /// This function is used to get current hashring and update current hashring
    /// If we set `must_fetch`, we will also update current hashring
    pub async fn get_ring(&self, must_fetch: bool) -> DatenLordResult<Ring<Node>> {}
}
```

The definition of Inner will be rather monolithic and the exposed methods are basically single tasks.
- load/save/get/update are exposed utility methods to manipulate etcd configuration/local configuration.
- watch_xxx() is a task to observe a key event.
- write_task() is a task for the master to periodically watch for events and write them to etcd.
- The remaining methods are encapsulated for ease of reuse.

### DIFF
--- a/src/storage/cache_proxy/cluster_manager.rs
+++ b/src/storage/cache_proxy/cluster_manager.rs
@@ -1,0 +1,1067 @@
+//! The utilities of distribute cache cluster management
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::Context;
+use arc_swap::{ArcSwap, ArcSwapOption};
+use futures::StreamExt;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, warn};
+
+use crate::async_fuse::memfs::kv_engine::etcd_impl::Session;
+use crate::async_fuse::memfs::kv_engine::{KVEngine, SetOption};
+use crate::async_fuse::memfs::kv_engine::{KVEngineType, KeyType, ValueType};
+use crate::common::error::{DatenLordError, DatenLordResult};
+
+use super::node::{MasterNodeInfo, Node, NodeStatus};
+use super::ring::Ring;
+
+/// The timeout for current node's session
+const SESSION_TIMEOUT_SEC: u64 = 10;
+
+/// This struct is used to interact with etcd server and manager the cluster.
+#[derive(Debug, Clone)]
+#[allow(dead_code)]
+pub struct ClusterManager {
+    /// Inner struct
+    inner: Arc<ClusterManagerInner>,
+}
+
+#[allow(dead_code)]
+impl ClusterManager {
+    /// Create new cluster manager
+    pub fn new(kv_engine: Arc<KVEngineType>, node: Node) -> Self {
+        let inner = Arc::new(ClusterManagerInner::new(kv_engine, node));
+        Self { inner }
+    }
+
+    /// Get current node
+    #[must_use]
+    pub fn get_node(&self) -> Node {
+        self.inner.get_node()
+    }
+
+    /// Get current hashring
+    pub async fn get_ring(&self) -> DatenLordResult<Ring<Node>> {
+        self.inner.get_ring(false).await
+    }
+
+    /// Get all nodes
+    pub async fn get_nodes(&self) -> DatenLordResult<Vec<Node>> {
+        self.inner.get_nodes(false).await
+    }
+
+    /// Run the cluster manager as state machine
+    ///
+    /// We need to perpare current node info, current node list and generate hash ring info
+    pub async fn run(&self, token: CancellationToken) -> DatenLordResult<()> {
+        loop {
+            // Check the shutdown flag, if current flag is set, we need return to the caller
+            if self.inner.shutdown_flag.load(Ordering::Acquire) {
+                return Ok(());
+            }
+
+            // 0. Prepare for the session
+            self.inner.prepare().await?;
+
+            // 1. Init cluster manager and register to etcd
+            let mut current_node_info = self.inner.get_node();
+            debug!(
+                "Cluster manager start to run in node: {:?} status: {:?}",
+                current_node_info.ip(),
+                current_node_info.status()
+            );
+            while self.inner.register().await.is_err() {
+                warn!("Failed to register node, retry in 5s");
+                // Try to register node to etcd
+                tokio::time::sleep(Duration::from_secs(5)).await;
+            }
+
+            // 2. Do campaign
+            let (campaign_status, campaign_val) = self.inner.do_campaign().await?;
+            debug!(
+                "Node: {:?} Campaign status: {:?} val: {:?}",
+                current_node_info.ip(),
+                campaign_status,
+                campaign_val
+            );
+            if campaign_status {
+                // Update node status to Master
+                current_node_info.set_status(NodeStatus::Master);
+                self.inner.update_node(current_node_info);
+            }
+
+            // 3. Serve as normal status
+            let current_node_info = self.inner.get_node();
+            match current_node_info.status() {
+                // Serve as slave node
+                NodeStatus::Slave => match self.do_slave_tasks(token.clone()).await {
+                    Ok(()) => {
+                        debug!(
+                            "Current node {:?} status: {:?} will change to campaign",
+                            current_node_info.ip(),
+                            current_node_info.status()
+                        );
+                    }
+                    Err(e) => {
+                        warn!("Failed to do slave tasks: {:?}", e);
+                    }
+                },
+                // Serve as master node
+                NodeStatus::Master => match self.do_master_tasks(token.clone()).await {
+                    Ok(()) => {
+                        debug!(
+                            "Current node {:?} status: {:?} will change to campaign",
+                            current_node_info.ip(),
+                            current_node_info.status()
+                        );
+                    }
+                    Err(e) => {
+                        warn!("Failed to do master tasks: {:?}", e);
+                    }
+                },
+            }
+        }
+    }
+
+    /// Do master tasks
+    ///
+    /// 1. Master node will watch the node list update, and update the ring
+    /// 2. Master will check self
+    async fn do_master_tasks(&self, token: CancellationToken) -> DatenLordResult<()> {
+        let current_node_info = self.inner.get_node();
+        debug!(
+            "do_master_tasks: {:?} will watch the node list update, and update the ring",
+            current_node_info.ip()
+        );
+
+        // Keep alive master key
+        // Check current status
+        if current_node_info.status() != NodeStatus::Master {
+            // If the node status is not master, clean up master tasks and return
+            warn!(
+                "Current node {:?} status is not master, return to endpoint",
+                current_node_info.ip()
+            );
+
+            return Ok(());
+        }
+
+        // Do watch node list update task
+        // This task will block until the master status changed
+        self.inner.watch_nodes(token).await.with_context(|| {
+            format!(
+                "Current node {:?} Failed to watch the node list update, and update the ring",
+                current_node_info.ip()
+            )
+        })?;
+
+        Ok(())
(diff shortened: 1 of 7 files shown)


src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+
+    /// Get all nodes
+    pub async fn get_nodes(&self) -> DatenLordResult<Vec<Node>> {
+        self.inner.get_nodes(false).await
+    }
+
+    ///  Stop the cluster manager
+    pub fn stop(&self) {
+        // In this step, we just to clean session and force down the cluster manager
+        self.inner.clean_sessions();
```
frostyplanet, line 57: How to foce shutdown?  after clean_session, run() will continue in the loop.

src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+            // 1. Init cluster manager
+            debug!(
+                "Cluster manager start to run in node: {:?}",
+                self.inner.get_node().ip()
+            );
+            let mut current_node_info = self.inner.get_node();
+            // Next step is to register the node
+            current_node_info.set_status(NodeStatus::Registering);
+            self.inner.update_node(current_node_info);
+            self.inner.update_node_in_cluster().await?;
```
frostyplanet, line 77: set_status(), update_node(), update_node_in_cluster() always call together? 
should be merge into one

src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+            .kv_engine
+            .create_session(SESSION_TIMEOUT_SEC)
+            .await
+            .with_context(|| format!("{} failed to create session", self.node.load().ip()))?;
+        self.node_session.store(Some(session));
+        Ok(())
+    }
+
+    /// Update node info
+    pub async fn update_node_in_cluster(&self) -> DatenLordResult<()> {
```
frostyplanet, line 265: If not exposed outside this file,  do not use pub

src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+                current_node_info.ip(),
+                current_node_info.status()
+            );
+            while self.inner.register().await.is_err() {
+                warn!("Failed to register node, retry in 5s");
+                // Try to register node to etcd
+                tokio::time::sleep(Duration::from_secs(5)).await;
+            }
+            // Update node status to Registering
+            current_node_info.set_status(NodeStatus::Slave);
```
frostyplanet, line 92: "set_status Slave" should be put after do_compaign()

frostyplanet (changes requested):

src/storage/cache_proxy/cluster_manager.rs
frostyplanet: `set_status(NodeStatus::Registering);` may be unnecessary

src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+                }),
+            )
+            .await
+            .with_context(|| "Failed to register node to etcd".to_owned())?;
+
+        Ok(())
+    }
+
+    /// Register current node to etcd and keep alive
+    pub async fn register(&self) -> DatenLordResult<()> {
```
frostyplanet, line 296: the loginc of register() and update_node_in_cluster() is exactly the same ?

src/storage/cache_proxy/cluster_manager.rs
```diff
@@ -0,0 +1,1096 @@
...
+
+    /// Do campaign
+    ///
+    /// Try to campaign master, will return status and master value
+    pub async fn do_campaign(&self) -> DatenLordResult<(bool, String)> {
+        let node = self.get_node();
+        let hash_ring = self.get_ring(false).await?;
+
+        let node_session = self.node_session.load();
+        let session = if let Some(session) = node_session.as_ref().cloned() {
```
frostyplanet, line 339: the code is long and repeat many times  (session.load & validation & warnning log),  and repeat many times ,  should be put in the function

src/storage/cache_proxy/cluster_manager.rs
frostyplanet: Master does not need to know each node's status.  Thus removing `status` from Node struct, the code will be much simpler

frostyplanet (changes requested):

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: The shutdown here is for test cases, if I need to force a shutdown I tend to set a flag and wait until the clean session is done loop will check the status of this flag.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: Yes, the values implemented here are to make it easier to understand the process, and I've removed duplicate steps.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: Stop `watch_xxx` task by channel event.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: However, NodeStatus is still required for local processing, and here I still made a reservation to facilitate the handling of role-related logic, and skipped NodeStatus during sequence talk to reduce the consistency maintenance of role status.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: I have removed this function.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: Thanks, current node status is modified to `Slave` as default value, this line will be removed.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: This was a legacy issue that has now been fixed.

src/storage/cache_proxy/cluster_manager.rs
IRONICBo: At this stage, NodeStatus may indeed not be necessary.

frostyplanet:
Need rebase

frostyplanet (approved):

frostyplanet:
LGTM

PR #529, opened 2024-07-19
Victor gsoc 2024

vicluo96:
Added xline_impl mod as a counterpart for etcd_impl

### DIFF
--- a/scripts/setup/start_local_node.sh
+++ b/scripts/setup/start_local_node.sh
@@ -12,7 +12,8 @@
 BUILD_FLAGS=$1
 
 . scripts/setup/config.sh
-. scripts/setup/setup_etcd.sh
+#. scripts/setup/setup_etcd.sh
+#. scripts/setup/setup_xline.sh
 
 if mount | grep -q "$DATENLORD_LOCAL_BIND_DIR"; then
     echo "$DATENLORD_LOCAL_BIND_DIR is mounted. Unmounting now."
--- a/src/async_fuse/memfs/kv_engine/xline_impl.rs
+++ b/src/async_fuse/memfs/kv_engine/xline_impl.rs
@@ -1,0 +1,365 @@
+use std::collections::HashMap;
+use std::fmt::Debug;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use datenlord::metrics::KV_METRICS;
+use anyhow::Context;
+use xline_client::{
+    types::{kv::{
+        Compare, CompareResult, DeleteRangeRequest, PutRequest, RangeRequest,
+        TxnOp, TxnRequest, 
+    }, lock::{LockRequest, UnlockRequest},
+    lease::LeaseGrantRequest
+    },
+    Client, ClientOptions,
+};
+
+
+use super::{
+    check_ttl, conv_u64_sec_2_i64, fmt, DeleteOption, KVEngine, KeyType, KvVersion, LockKeyType,
+    MetaTxn, SetOption, ValueType,
+};
+//use crate::common::error::{Context
+use crate::common::error::DatenLordResult;
+
+#[derive(Clone)]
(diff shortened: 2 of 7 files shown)


frostyplanet:
ci error might need look into    ` error: failed to run custom build command for `curp v0.1.0`

scripts/setup/start_local_node.sh
```diff
@@ -12,7 +12,8 @@
 BUILD_FLAGS=$1
 
 . scripts/setup/config.sh
-. scripts/setup/setup_etcd.sh
+#. scripts/setup/setup_etcd.sh
+#. scripts/setup/setup_xline.sh
```
frostyplanet, line 16: Uncomment  `. scripts/setup/setup_xline.sh `,  CI might depend on it.

IRONICBo:
image
Please check the compilation status.

src/async_fuse/memfs/kv_engine/xline_impl.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    /// Delete the kv pair by the key.
+    async fn delete(
+        &self,
+        key: &KeyType,
+        _option: Option<DeleteOption>,
+    ) -> DatenLordResult<Option<ValueType>> {
+        let _timer = KV_METRICS.start_kv_operation_timer("delete");
+        let mut delete_req = DeleteRangeRequest::new(key.to_string_key());
+        let _option = match _option {
```
IRONICBo, line 207: Remove `_` prefix in clippy test.

src/async_fuse/memfs/kv_engine/xline_impl.rs
```diff
@@ -0,0 +1,365 @@
...
+                                }
+                            })
+                            .collect::<Vec<TxnOp>>(),
+                    ),
+            )
+            .await
+            .with_context(|| "failed to do txn operation at `MetaTxn::commit`".to_owned())?;
+        Ok(resp.succeeded)
+        }
+}
```
IRONICBo, line 365: Please add `UT`s for xline meta engine.

src/async_fuse/test/test_util.rs
```diff
@@ -60,8 +60,10 @@ fn test_storage_config(is_s3: bool) -> StorageConfig {
 
 async fn run_fs(mount_point: &Path, is_s3: bool, token: CancellationToken) -> anyhow::Result<()> {
     let storage_config = test_storage_config(is_s3);
-    let kv_engine: Arc<memfs::kv_engine::etcd_impl::EtcdKVEngine> =
-        Arc::new(KVEngineType::new(vec![TEST_ETCD_ENDPOINT.to_owned()]).await?);
+    // let kv_engine: Arc<memfs::kv_engine::etcd_impl::EtcdKVEngine> =
+    //     Arc::new(KVEngineType::new(vec![TEST_ETCD_ENDPOINT.to_owned()]).await?);
+    let kv_engine: Arc<memfs::kv_engine::xline_impl::XlineKVEngine> =
+    Arc::new(KVEngineType::new(vec![TEST_ETCD_ENDPOINT.to_owned()]).await?);
```
IRONICBo, line 66: As we mentioned earlier, it may be better to use enum wrapping, plus the type of meta engine can be specified in config.

src/async_fuse/memfs/kv_engine/xline_impl.rs
vicluo96: Hi @IRONICBo, since xline use request as the parameter  in `delete` and `set` function, so the `_option` will not be used. Therefore, remove the underscore followed by `option` would cause compile error. I am thinking whether I should use request as parameter type instead of using option.

IRONICBo: @vicluo96 Yes, the options here are mainly adapted to etcd requests, maybe consider adapting the Option type to xline requests.

PR #530, opened 2024-08-02
fix: current block flush task failed in ut and cause inconsistent behavior in cargo test

IRONICBo:
Currently mem cache and mem storage encounter write backend failures during concurrency testing unit tests, i.e., when managing **write_back_one_block** here, the sender and receiver are affected by the global TASK_MANAGER, and there's no way to receive the task properly, causing the Brushing the backend (mem storage) fails, which is **inconsistent** with the expected behavior.

The problem occurs when it is blocked or a send error occurs, resulting in a failure to write to the backend. ~~The temporary modification here is to cancel the task that submits the block flush to ensure that the test point can pass.~~

Current block flush task failed in ut and cause inconsistent behavior in cargo test, reference issue is #531.
    
Because we want to use singleton task manager, this manager will create a task in current runtime context, it is isolated in tokio::test.

Reference issue: https://github.com/datenlord/datenlord/issues/531

### DIFF
--- a/src/common/task_manager/tests.rs
+++ b/src/common/task_manager/tests.rs
@@ -14,20 +14,27 @@ use crate::common::task_manager::manager::SpawnError;
 use crate::common::task_manager::task::{TaskName, EDGES};
 use crate::common::task_manager::{wait_for_shutdown, TASK_MANAGER};
 
-#[tokio::test]
-async fn test_dependency_graph() {
+#[test]
+fn test_dependency_graph() {
     let task_manager = TaskManager::new();
-    let edges_dumped: HashSet<_> = task_manager.edges().await.into_iter().collect();
-    let edges_expected: HashSet<_> = EDGES.into_iter().collect();
+    let test_runtime = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
+        .enable_all()
+        .build()
+        .unwrap();
 
-    assert_eq!(edges_dumped, edges_expected);
+    test_runtime.block_on(async {
+        let edges_dumped: HashSet<_> = task_manager.edges().await.into_iter().collect();
+        let edges_expected: HashSet<_> = EDGES.into_iter().collect();
+        assert_eq!(edges_dumped, edges_expected);
 
-    let predecessor_counts_dumped = task_manager.predecessor_counts().await;
+        let predecessor_counts_dumped = task_manager.predecessor_counts().await;
 
-    let mut predecessor_counts_expected = EDGES.iter().counts_by(|&(_, task_name)| task_name);
-    predecessor_counts_expected.insert(TaskName::Root, 0);
+        let mut predecessor_counts_expected = EDGES.iter().counts_by(|&(_, task_name)| task_name);
+        predecessor_counts_expected.insert(TaskName::Root, 0);
 
-    assert_eq!(predecessor_counts_dumped, predecessor_counts_expected);
+        assert_eq!(predecessor_counts_dumped, predecessor_counts_expected);
+    });
 }
 
 /// Create a future, that waits for a signal from `token`, then sends a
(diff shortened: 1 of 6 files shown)


src/storage/memory_cache/tests.rs
```diff
@@ -372,6 +372,8 @@ async fn test_store_write_back() {
     let mut block = Block::new_zeroed(BLOCK_SIZE_IN_BYTES);
     block.set_dirty(true);
     cache.store(0, 0, block).await.unwrap();
+    // Wait for the write back task
+    tokio::time::sleep(Duration::from_millis(200)).await;
```
frostyplanet, line 376: Does cache.flush ensure wait for writeback finished ? If it's true,  it's better to use flush than sleep for a time (since it depends on test env)

frostyplanet:
Need more explaination in the commit message.

IRONICBo:
tests:
```bash
test storage::storage_manager::test::common::test_remove ... ok
test storage::storage_manager::test::common::test_truncate ... ok
test storage::storage_manager::test::common::test_truncate_remove ... ok
test storage::storage_manager::test::common::test_zero_size_read_write ... ok
test storage::storage_manager::test::concurrency::test_concurrency_aligned ... ok
test storage::storage_manager::test::concurrency::test_concurrency_unaligned ... ok
test storage::storage_manager::test::latency::test_write_back_flush_latency ... ok
test storage::storage_manager::test::latency::test_write_back_latency ... ok
test storage::storage_manager::test::latency::test_write_through_latency ... ok

test result: ok. 83 passed; 0 failed; 0 ignored; 0 measured; 85 filtered out; finished in 29.71s
```

src/storage/memory_cache/tests.rs
```diff
@@ -152,203 +182,225 @@ async fn prepare_data_for_evict() -> (Arc<MemoryStorage>, Arc<MemoryCacheType>)
...
-    // LRU in cache: (0, 1) -> (0, 2) -> (0, 3) -> (0, 0)
-    // Insert a block, and (0, 1) will be evicted
-    assert!(!backend.contains(0, 1));
-    let block = Block::from_slice(BLOCK_SIZE_IN_BYTES, BLOCK_CONTENT);
-    cache.store(1, 0, block).await.unwrap();
-    let loaded = backend.load(0, 1).await.unwrap().unwrap();
-    assert_eq!(loaded.as_slice(), BLOCK_CONTENT);
+#[rstest]
+fn test_touch_by_load() {
+    shared_runtime_test(async {
```
frostyplanet, line 202: task manager have a runtime, do not need shared_runtime_test anymore

src/common/task_manager/tests.rs
```diff
@@ -14,20 +14,27 @@ use crate::common::task_manager::manager::SpawnError;
 use crate::common::task_manager::task::{TaskName, EDGES};
 use crate::common::task_manager::{wait_for_shutdown, TASK_MANAGER};
 
-#[tokio::test]
-async fn test_dependency_graph() {
+#[test]
```
frostyplanet, line 17: Can you revert the changes in test to see what happend?

frostyplanet (changes requested):

src/storage/memory_cache/tests.rs
IRONICBo: OK, I have revert this commit, it is not necessary here by shared runtime.

src/common/task_manager/tests.rs
IRONICBo: tokio::test will start a new tokio runtime and drop the temp task manager in current runtime, it is not allowed in async context:
```bash
.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.37.0/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
```

frostyplanet (approved):
LGTM

frostyplanet (approved):

PR #536, opened 2024-08-23
fix: new storage cache manager will cause unexcepted cache clean.

IRONICBo:
# Bug

The current file handling mechanism aggressively removes cache, but we want to retain this cache when continuous read/write operations are occurring.

# Fix

Update the cache with a versioning system. The LRU (Least Recently Used) policy will automatically clean up outdated cache entries when memory is insufficient.

Additionally, we can add a `file_version` to each block to quickly identify and discard outdated blocks in access function. The current implementation does not utilize this approach.

## TODO

1. Set version to cache value, write task can access outdate block and write this cache.
2. write 256k & write 4k mechaism for block version and block cache.

benchmark:
1. old policy:
```bash
python3 test_reader.py -n 10
Average read time: 0.21317 s
Maximum read time: 0.24317 s
Minimum read time: 0.18317 s
Median read time: 0.22317 s
```
3. new policy with version:
```bash
python3 test_reader.py -n 10
Average read time: 0.04402 s
Maximum read time: 0.24390 s
Minimum read time: 0.01637 s
Median read time: 0.02104 s
```

### DIFF
--- a/src/new_storage/handle/reader.rs
+++ b/src/new_storage/handle/reader.rs
@@ -46,11 +46,9 @@ impl Reader {
         let key = CacheKey {
             ino: self.ino,
             block_id,
-            // Check current block is invalid or not.
-            version,
         };
         let cache = self.cache.lock();
-        cache.fetch(&key)
+        cache.fetch(&key, version)
     }
 
     /// Fetch the block from the backend storage system.
@@ -62,7 +60,6 @@ impl Reader {
         let key = CacheKey {
             ino: self.ino,
             block_id,
-            version,
         };
         let content = {
             let mut buf = vec![0; self.block_size];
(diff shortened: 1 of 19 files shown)


src/new_storage/handle/reader.rs
```diff
@@ -41,40 +38,36 @@ impl Reader {
...
     }
 
     /// Try fetch the block from `MemoryCache`.
-    fn fetch_block_from_cache(&self, block_id: u64) -> Option<Arc<RwLock<Block>>> {
+    fn fetch_block_from_cache(&self, block_id: u64, version: u64) -> Option<Arc<RwLock<Block>>> {
         let key = CacheKey {
             ino: self.ino,
             block_id,
+            // Check current block is invalid or not.
+            version,
```
frostyplanet, line 50: It might be better not change the cache key, put version into cache item instead.
when old version found from cache, the cache item can be deleted immediately.

IRONICBo: Ok, now that the structure of the Map is HashMap<K, Arc<RwLock<Block>>>, try to associate the version with the block and remove the version field from the cache key, but it may slightly affect the retrieval speed.

IRONICBo:
Dismiss `test_open_file_permission` for metadata.

IRONICBo:
Moved to #549.

Issue #499, opened 2024-03-12
Some Misc Issues in Rank 1


Reto911:
- [ ] Some functions in the Session use trait objects, and we can make them generic to reduce the performance loss caused by dynamic dispatch. However, considering that the `async_trait` currently in use already uses dynamic dispatch extensively, this may not have a significant impact.
- [ ] When negotiating the protocol, there are still two cases that have not been handled. Before implementing dynamic protocol negotiation, at least an error should be reported and the program should exit in these two cases:

   1. The kernel supports a version lower than the version specified by the compilation options of the current Datenlord (for example, the kernel supporting 7.11, while the current software uses 7.23).
   2. The major version number is greater than 7. This is not currently possible because the highest major version number for FUSE is still 7.

- [ ] Re-evaluate the `can_convert` function in `AsIoVec`. If it is not necessary, remove it.
- [ ] Rewrite all paths in macros as absolute paths.
- [ ] The implementation of `Rename` is too verbose. Optimize it in terms of code style.
- [ ] #371
- [ ] Implement `convert_to_fuse_attr` as a `From` trait.

msohailx:
hey @Reto911 please assign this issue to me

plon-Susk7:
hey @Reto911 , I am new to this repo and would love to contribute. Shall I work on this issue?

Reto911:
> hey @Reto911 , I am new to this repo and would love to contribute. Shall I work on this issue?

Thank you for your interest, I am sorry for replying so late. I think you can, but I am not sure about that now, you can try to contact with other active collaborators.

PR #550, opened 2024-11-22
refactor: implement convert_to_fuse_attr as a From trait

tkkkk2:
This pr implements convert_to_fuse_attr as a From trait as stated in #499.

### DIFF
--- a/src/async_fuse/memfs/fs_util.rs
+++ b/src/async_fuse/memfs/fs_util.rs
@@ -273,6 +273,35 @@ impl Default for FileAttr {
     }
 }
 
+impl From<FileAttr> for FuseAttr {
+    fn from(attr: FileAttr) -> Self {
+        let (a_time_secs, a_time_nanos) = time_from_system_time(&attr.atime);
+        let (m_time_secs, m_time_nanos) = time_from_system_time(&attr.mtime);
+        let (c_time_secs, c_time_nanos) = time_from_system_time(&attr.ctime);
+
+        FuseAttr {
+            ino: attr.ino,
+            size: attr.size,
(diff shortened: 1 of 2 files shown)


src/async_fuse/memfs/fs_util.rs
```diff
@@ -273,6 +273,35 @@ impl Default for FileAttr {
...
+            ino: attr.ino,
+            size: attr.size,
+            blocks: attr.blocks,
+            atime: a_time_secs,
+            mtime: m_time_secs,
+            ctime: c_time_secs,
+            atimensec: a_time_nanos,
+            mtimensec: m_time_nanos,
+            ctimensec: c_time_nanos,
+            mode: crate::async_fuse::util::mode_from_kind_and_perm(attr.kind, attr.perm),
```
IRONICBo, line 292: Suggest replacing with `util::mode_from_kind_and_perm`.

tkkkk2: Thank you for your feedback. Since it already imports module `use crate::common::util`, I don't think this will work directly. Could you please make some suggestions?

IRONICBo: You can import this crate at the top of this file `use xxx::util as fuse_util` or directly import this help function.

tkkkk2: OK, I have fixed it. Please review the changes.

IRONICBo (approved):
LGTM

IRONICBo:
@frostyplanet Please take a look.

frostyplanet (approved):
LGTM

PR #554, opened 2024-11-28
feat: split fuse api for python sdk

IRONICBo:
# Background

At present, the storage interface exposed to the outside world is POSIX (fuse), in order to consider the user space to complete the transmission of storage data, consider providing a set of sdk for the business layer to efficiently read the data stored in the back-end without going through the kernel, you can refer to the form of the deployment of the following two main forms of discussion. 

# Core Modification

1. Add c SDK for datenlordfs.

# TODO

### DIFF
--- a/sdk/c/examples/test_datenlord_sdk.c
+++ b/sdk/c/examples/test_datenlord_sdk.c
@@ -1,0 +1,167 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include "datenlordsdk.h"
+
+#define FILE_SIZE_MB 20
+#define FILE_SIZE (FILE_SIZE_MB * 1024 * 1024)
+#define READ_BUFFER_SIZE FILE_SIZE
+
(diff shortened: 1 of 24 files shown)


sdk/python/examples/bench_fuse_read.c
```diff
@@ -0,0 +1,90 @@
+#include <stdio.h>
```
frostyplanet, line 1: This file should in on C sdk dir

sdk/c/examples/test_datenlord_sdk.c
```diff
@@ -0,0 +1,167 @@
+#include <stdio.h>
+#include <stdlib.h>
```
frostyplanet, line 2: should be in c sdk dir

frostyplanet (changes requested):

sdk/python/examples/bench_fuse_read.c
IRONICBo: Ok, moved.

sdk/c/examples/test_datenlord_sdk.c
IRONICBo: Ok, moved.

PR #556, opened 2024-11-28
feat: introduce distribute kvcache.

IRONICBo:
# Background

The current phase of distribute kv cache has been done in the upper layer of the upper layer of the design, you can see the alignment of the distribute kv cache part of the core changes in the processing of the radix tree, with reference to the original distribute cache architecture, the master needs to deal with the core of the metadata The master needs to handle the core metadata service, i.e., the radix tree part, and the remaining slave nodes participate in the actual kvcache memory storage.

# Design

For master node:

1. GlobalIndexManager
Used by the master to store the radix state in the cluster, provides read and write interfaces, the internal structure mainly includes:
index: generic trie read/write interface definition, supports concurrency
id_allocator: Used to assign a unique blockid for the cluster.
In addition, GlobalIndexManager supports Serial and Deserial interfaces, and puts the current index into kvengine.

2. KVBlockManager
Used to store kvblock, and the original distribution cache in the mem cache is similar, here do not go into details, direct reuse.

3. GlobalIndexHandler(RPC)
Used to handle server-side read and write requests to the radix tree, internal processing of read and write requests for radix & index.

4.  KVBlockHandler(RPC)
Used to handle server-side api read and write requests, basically reuse the previous logic, not to repeat here.


Foreverhighness:
There is a potential race condition or message losing between `try_recv` in `PacketsKeeper::take_task` and `add_task` in `RpcClientConnectionInner::send_packet`.

**Problem** 
Execution order sensitivity between `try_recv` and `add_task`:
- ✅ Safe path: `add_task` → `try_recv` (normal operation)
- ❌ Failure path: `try_recv` → `add_task` (client deadlock)

**Reproduction** 
It can be reproduced by insert 100ms delay before `add_task`.
```rust
pub async fn send_packet(&self, mut req_packet: P) -> Result<(), RpcError> {
    // ...
    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
    self.packets_keeper.add_task(req_packet)?;
    // ...
}
```

PR #558, opened 2024-11-29
Fix deferred deletion behavior

frostyplanet:
We implement deferred deletion only for local node, drop the support between multi-nodes.
 (in multi-node enviroment there might be possible to leave file lookup_count>0 which cannot never be deleted once related node is down)

1. According to posix behavior, opened file get unlink is deleted from directory immediately. File will be deleted on close.
    
2. Remove previous lookup_count logic which is not reliable (with VFS page cache behavior),  might lead to unexpect panic on file deletion.

NOTE: e2e test still rely on deferred deletion feature.

Additional change:  cleanup log format and simplify FileHandle code


IRONICBo (approved):
LGTM

PR #565, opened 2025-06-06
feat: introduce kvcache client in rust.

IRONICBo:
# Target

In this PR, we want to introduce the distribute kvcache client in rust, and support to read and write metadata and data by client, this PR contains:
- Distribute cache client.
- Read and wirte operation.

# Design

The above module corresponds to the distribiute cache manager:

##### KVBlock

```rs
/// `KVBlock` instance
///
/// |--------`KVBlock`-----| => single kv block data
/// |cache1|cache2|cache3| => contains kv cache data = cache1, cache2, cache3
/// |0-----|1-----|2-----| => prefix = 0, 1, 2
#[derive(Debug, Clone)]
pub struct KVBlock {
    /// The block id
    pub block_id: u64,
    /// The block size
    pub block_size: u64,
    /// The block data, contains one or more kv cache data
    pub data: Vec<u8>,
}

```

- KVBlock contains many kvcache in one block.

##### LocalBlockCache

```rs
/// Local block cache
#[derive(Debug, Clone)]
struct LocalBlockCache<K>
where
    K: num::Num + Eq + Send + Sync + Clone + fmt::Debug + 'static,
{
    /// The block data
    block_cache: KVBlock,
    /// The block metas
    block_metas: Vec<KVCacheMeta<K>>,
    /// The radix tree for block metas, support partial match
    /// Use token ids as the key, and the kv cache meta as the value
    /// u32 size is enough for the token id in tokenized.
    block_metas_tree: Trie<Vec<K>, KVCacheMeta<K>>,
}
```
- Client will collect kvcache and write kvcache in one op

##### Client

```rs
/// The distribute cache client
#[derive(Debug, Clone)]
pub struct DistributeKVCacheClient<K>
where
    K: num::Num + Eq + Send + Sync + Clone + fmt::Debug + Serialize + DeserializeOwned + 'static,
{
    /// The distribute cache inner
    inner: DistributeKVCacheClientInner<K>,
    /// Single block cache, used to collect the block data from the infer side.
    block_cache: Arc<Mutex<LocalBlockCache<K>>>,
}

#[derive(Debug, Clone)]
pub struct DistributeKVCacheClientInner<K>
where
    K: num::Num + Eq + Send + Sync + Clone + fmt::Debug + Serialize + DeserializeOwned + 'static,
{
    /// The cluster manager, only used to watch hashring changes
    cluster_manager: Arc<ClusterManager>,
    /// The rpc client cache
    /// TODO: Change to trait
    rpc_client_cache: Arc<Mutex<HashMap<String, Arc<RpcClient<KVCacheRequestTask<K>>>>>>,
    /// Block size
    block_size: u64,
}
```

- Client will export read and write op in this struct.

### DIFF
--- a/src/distribute_kv_cache/kvclient.rs
+++ b/src/distribute_kv_cache/kvclient.rs
@@ -1,0 +1,1173 @@
+use core::fmt;
+use std::{collections::HashMap, sync::Arc, time::Duration};
+
+use clippy_utilities::Cast;
+use radix_trie::Trie;
+use tokio::sync::Mutex;
+use tracing::{debug, error, info, warn};
+
+use crate::{
+    async_fuse::util::usize_to_u64,
+    common::{
+        error::{DatenLordError, DatenLordResult},
+        task_manager::{TaskName, TASK_MANAGER},
+    },
+    connect_timeout,
+    distribute_kv_cache::rpc::message,
+};
+
+use super::{
+    cluster::cluster_manager::ClusterManager,
+    rpc::{
+        client::RpcClient,
+        common::ClientTimeoutOptions,
+        message::{
+            KVBlockBatchPutRequest, KVBlockGetRequest, KVBlockPutRequest, KVCacheIdAllocateRequest,
+            KVCacheIndexBatchInsertRequest, KVCacheIndexInsertRequest, KVCacheIndexMatchRequest,
+            KVCacheIndexRemoveRequest, KVCachePacket, KVCacheRequest, KVCacheResponse, ReqType,
+        },
+        utils::u64_to_usize,
+    },
+};
+
+/// Duration to check the rpc client cache
+const RPC_CLIENT_CACHE_CHECK_DURATION: u64 = 60;
+
+/// Unused kv block id
+const UNUSED_KV_BLOCK_ID: u64 = 0;
+
+/// KVBlock instance
+///
+/// |--------KVBlock-----| => single kv block data
+/// |cache1|cache2|cache3| => contains kv cache data = cache1, cache2, cache3
+/// |0-----|1-----|2-----| => prefix = 0, 1, 2
+///
+/// TODO: Hold the kv cache and fetch from local cache
+#[derive(Debug, Clone)]
+pub struct KVBlock {
+    /// The block id
+    pub block_id: u64,
+    /// The block size
+    pub block_size: u64,
+    /// The block data, contains one or more kv cache data
+    pub data: Vec<u8>,
+}
+
+/// KVCache instance metadata
+///
+/// Set current kv cache metadata info to single kv block
+#[derive(Debug, Clone)]
+pub struct KVCacheMeta<K>
+where
+    K: num::Num + Eq + Send + Sync + Clone + fmt::Debug + 'static,
+{
+    /// The kv block id
+    pub block_id: u64,
+    /// The kv cache offset
+    pub offset: u64,
+    /// The kv cache size
+    pub size: u64,
+    /// The kv cache prefix
+    pub prefix: Vec<K>,
+}
+
+impl<K> Default for KVCacheMeta<K>
+where
+    K: num::Num + Eq + Send + Sync + Clone + fmt::Debug + 'static,
+{
+    fn default() -> Self {
+        Self {
+            block_id: UNUSED_KV_BLOCK_ID,
+            offset: 0,
+            size: 0,
+            prefix: Vec::new(),
+        }
+    }
+}
+
+/// Local block cache
+#[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


src/distribute_kv_cache/kvclient.rs
```diff
@@ -0,0 +1,970 @@
...
+
+        let next_offset = block_cache.get_next_offset();
+        let kv_cache_meta = KVCacheMeta {
+            block_id: current_block_id,
+            offset: next_offset,
+            size: data.len().cast(),
+            prefix: prefix.clone(),
+        };
+
+        match block_cache.insert(kv_cache_meta, &data) {
```
rogercloud, line 449: block_cache's `clear` and `insert` always called together, why don't you merge them?

src/distribute_kv_cache/kvclient.rs
```diff
@@ -0,0 +1,970 @@
...
+                    .inner
+                    .cluster_manager
+                    .get_node_remote(kv_block.block_id.to_string())
+                    .await?;
+
+                let addr = format!("{}:{}", node.ip(), node.port());
+                self.inner.put_blocks(addr.clone(), vec![kv_block]).await?;
+
+                // If ok, insert the indexes to the distribute cache
+                self.inner.insert_indexes(kv_cache_metas, addr).await?;
```
rogercloud, line 474: If the index insert fails, will you revert the block insert operation? In another word, will you gc useless data block?

src/distribute_kv_cache/kvclient.rs
```diff
@@ -0,0 +1,970 @@
...
+                Ok(current_ring) => {
+                    if !current_ring.is_empty() && current_ring.version() != 0 {
+                        info!("Current ring is stable: {:?}", current_ring);
+                        break;
+                    }
+                }
+                Err(err) => {
+                    warn!("Failed to get ring: {:?}", err);
+                }
+            }
```
rogercloud, line 559: Wait for a while if failed?

rogercloud (commented):
waht's the purpose of rpc client cache?

IRONICBo:
> waht's the purpose of rpc client cache?

The inference engine may write very small blocks, and we want to be able to collect these smaller blocks on the SDK and write them to the remote server in one go.

src/distribute_kv_cache/kvclient.rs
IRONICBo: That makes sense. I'll add it.

src/distribute_kv_cache/kvclient.rs
```diff
@@ -0,0 +1,970 @@
...
+                "Matched local kv cache meta from local cache: {:?}",
+                kv_cache_meta
+            );
+            kv_cache_meta
+        } else {
+            debug!("Failed to match local kv cache meta from local cache, try to get from the distribute cache");
+            // Empty meta
+            KVCacheMeta::default()
+        };
+        let local_prefix_len = local_kv_cache_meta.prefix.len().cast::<u64>();
```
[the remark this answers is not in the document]
IRONICBo: That makes sense. I'll split it.

src/distribute_kv_cache/kvclient.rs
```diff
@@ -0,0 +1,970 @@
...
+                // Return a empty data
+                (KVCacheMeta::default(), String::new())
+            }
+        };
+
+        let remote_prefix_len = kv_block_meta.prefix.len().cast::<u64>();
+        info!(
+            "Matched remote kv block meta: {:?} node_address: {:?}",
+            kv_block_meta, node_address
+        );
```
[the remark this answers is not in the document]
IRONICBo: That makes sense. I'll split it.

src/distribute_kv_cache/kvclient.rs
rogercloud: I prefer merging these two functions.

src/distribute_kv_cache/kvclient.rs
rogercloud: So it's a lazy GC by LRU. that makes sense.

src/distribute_kv_cache/kvclient.rs
IRONICBo: OK, changed to `clear_and_insert`.

IRONICBo: After confirmation, the clear should still be separated here. Only when insert encounters an error (after being filled) will it clear and reapply for a block ID. If there is still remaining space, there is no need to clear the block.

IRONICBo: Change `clear()` to `refresh_id_and_clear()`.

rogercloud (approved):

PR #567, opened 2025-06-16
feat: add integration test for kvcache.

IRONICBo:
In this PR, we want to add ut for kvcache client and server with simple request and response check here.


src/distribute_kv_cache/tests/kvcache.rs
```diff
@@ -0,0 +1,176 @@
+use std::sync::Once;
+
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{
+    filter, fmt::layer, layer::SubscriberExt, util::SubscriberInitExt, Layer,
+};
+
+/// Use for unit test to setup tracing
+#[allow(dead_code)]
```
rogercloud, line 9: why dead code?

src/distribute_kv_cache/tests/kvcache.rs
```diff
@@ -0,0 +1,176 @@
...
+static INIT: Once = Once::new();
+
+/// Set up once for tracing
+#[allow(dead_code)]
```
rogercloud, line 13: same as above

src/distribute_kv_cache/tests/kvcache.rs
```diff
@@ -0,0 +1,176 @@
...
+        let elapsed = start.elapsed();
+        println!(
+            "Speed: {:?}",
+            16777216.0 / 1024.0 / 1024.0 / elapsed.as_secs_f64()
+        );
+
+        // Test get 4 from the kv cache client, should return error
+        let prefix = vec![1, 2, 3, 4];
+        let (_, buf) = kvcacheclient.try_load(prefix.clone()).await.unwrap();
+        assert_eq!(buf.len(), 0);
```
rogercloud, line 167: also check error?

src/distribute_kv_cache/tests/rpc.rs
```diff
@@ -0,0 +1,383 @@
...
+        let rpc_client = RpcClient::<FileBlockRequestTask>::new(connect_stream, &timeout_options);
+        rpc_client.start_recv();
+
+        // Drop client
+        drop(rpc_client);
+
+        // Wait some time to ensure the client has dropped, wait for tcp read timeout. default is 20 second
+        time::sleep(Duration::from_secs(20)).await;
+
+        // Check if the server has detected the dropped connection
```
rogercloud, line 377: Where is the check?

src/distribute_kv_cache/tests/kvcache.rs
IRONICBo: Previously, there were some tools and methods that allowed lowering the level in tests. However, it is now possible to consider removing dead code.

src/distribute_kv_cache/tests/kvcache.rs
IRONICBo: Will be removed.

src/distribute_kv_cache/tests/kvcache.rs
IRONICBo: Make sense, I will add it.

PR #571, opened 2025-06-26
feat: add e2e test for distribute kvcache.

IRONICBo:
Test case:
1. check 3 node for kvcache server.
2. Kill 1 node and check availability.
3. Kill 2 node and check availability.


IRONICBo:
CI seems to be failed, I will check it later.

